Issue #21, opened 2023-08-10
Option to accept an access token instead of username/password


notself:
Is it possible to accept an access token directly instead of a username and password?

This would allow the app to work with the following Paperless setup, which I use:

- Authelia used to manage access to Paperless
- Nginx used as a reverse proxy with the `auth_request` module configured to use Authelia to authorize requests
- Paperless configured with `PAPERLESS_ENABLE_HTTP_REMOTE_USER=true` and `PAPERLESS_HTTP_REMOTE_USER_HEADER_NAME` set to the header configured in Authelia. In my case the default `HTTP_REMOTE_USER` value
- Nginx configured to bypass Authelia for Paperless `/api` endpoints

With this setup, it means that:
- User accounts are centralised behind Authelia's user store (in my case a local LDAP server)
- User accounts in Paperless are created automatically on first access
- Paperless accounts don't have a password set
- To access the Paperless API, we can use Django's Admin interface and create a token in `admin/authtoken/tokenproxy/` for an existing user

I see the app already allows to add custom headers but at the moment, even if we manually create an `Authorization: Token xxxx` header, the login flow is still there.

paulgessinger:
Hey @notself, yeah that's a good idea actually. 
In the end I really only save the token anyway, so one could just bypass the user/pw login flow by setting the token directly.

I'll look into this!

notself:
Amazing. Can't wait to try this app!

paulgessinger:
This has been on TestFlight for a while now. Did you get a chance to test it @notself?

pyrho:
Tested with Traefik + Tinyauth, works great thanks !

PR #26, opened 2023-10-09
feat: Add ability to login using an access token

RedcedarGeoff:
Allow a user to login with an access token by leaving the username field blank and placing the access token in the password field.

The user interaction/flow/UI for this as it is implemented now is far from ideal. If I could get some direction on how this interaction should/could be, I will make the changes.


paulgessinger:
Hey @RedcedarGeoff! Thanks for this PR! I'll give this a try on the weekend and then get back to you!

paulgessinger:
Hey @RedcedarGeoff I finally got a chance to check it out and test it on my end! Looks good to me and it's working.

I'm ok to merge this as is and follow this up with some UI, or you can change the PR to include UI for it if you want.

I guess in terms of UI I have two ideas:

1. Keep the basics of the UI as is, change "Password" to "Password / Token", and then put a `?` SF Symbol to the right. Tapping on that symbol opens a basic dialog explaining to leave the username empty for token login.
2. Use a switch or a picker to choose between password and token login, and then effectively just hide the username field and change the password placeholder text to "Token"

Let me know what you think!

RedcedarGeoff:
@paulgessinger I'm happy to make those UI changes. I'll try both suggestions and see what looks/feels right.

Based on issue #21, adding the `Authorization` manually could also work, but that the management of the `Authorization` header seems fairly deeply embedded in the logic, and would require some major refactoring in order to go that route?

paulgessinger:
You're right that explicitly setting the Authorization header would be relatively involved, I think.
To me handling this on the login screen seems like the better approach at this time.

paulgessinger:
@JamesRy96 I have an alternative implementation that's at the top of my list as soon as the iOS 18 update is ready.

paulgessinger:
An alternative to this is now available for testing in build 106 on TestFlight. I'm closing this PR.

@JamesRy96

PR #29, opened 2023-11-02
Update Localizable.strings

Christophe85:
Translate in French


Christophe85:
Great!! It looks like very good!

Thanks

PR #38, opened 2023-12-15
change delete-button text

LuegM:
Hey,
this is the first pull-request in my life. 😅
So please just let me know if something seems wrong.

**I have corrected the text of the delete button.**

There is also a problem on the iPad, which I don't know how to solve right now. 
I have also sent you an e-mail about this.

Best regards,
Michael


LuegM:
DeleteButton

paulgessinger:
Ha! That's indeed a bug! Thanks

Issue #59, opened 2024-02-12
Adding file via camera doesn't work?


phil-hudson:
Hey, I'm trying to add some new files into my paperless instance via the app.

After using the plus button, I'm not seeing the document added. Am I using it wrong in some way?

Please see repro attached

https://github.com/paulgessinger/swift-paperless/assets/1914950/9852d8a8-eec9-4dcf-a3bc-5be44fe7b6ff

paulgessinger:
Hey @phil-hudson,

After saving, the document creation sheet should pop up. Are you tapping the "Save" button in the bottom left?

phil-hudson:
Ah yep, it doesn't show in the video but I tap save, and then the camera is dismissed, but nothing happens. No error on log export too.

Nice work on this project btw!

evankrob:
Just to confirm. I have the same issue as Phil-Hudson

paulgessinger:
I'll add some logging to the scan workflow, then maybe we can figure this one out.

paulgessinger:
@evankrob @phil-hudson Can you check build 39 on TestFlight and send me the logs?

ftruzzi:
Hi @paulgessinger, I'm having the same issue. Here are the logs with build 39:

> [13/2/2024, 13:25] [SwiftPaperless] API sequence has reached end (nextPage is nil)
[13/2/2024, 13:25] [SwiftPaperless] API sequence has reached end (nextPage is nil)
[13/2/2024, 13:25] [SwiftPaperless] Fetch all store complete
[13/2/2024, 13:25] [SwiftPaperless] Document scanner receives scan
[13/2/2024, 13:25] [SwiftPaperless] Attempt to make PDF
[13/2/2024, 13:25] [SwiftPaperless] PDF conversion failure: Error Domain=swift_paperless.DocumentScannerView.DocumentScannerError Code=0
[13/2/2024, 13:25] [SwiftPaperless] Requesting log export

Thanks for your help, and nice app! :)

paulgessinger:
Ok so the failure seems to be that the PDF creation from the image source fails, which is unfortunate because it's relatively opaque and inside the framework.

phil-hudson:
@evankrob probably needs some debugging here: https://github.com/paulgessinger/swift-paperless/blob/97593e3fd483a8f3e35aa466572fec9ab175be7e/swift-paperless/Views/DocumentScannerView.swift#L55

paulgessinger:
[...]

PR #64, opened 2024-03-19
Fixed date formatter instead of locale dependant

PTST:
Fix for #59

At least on my phone the error was caused by my locale using "/" as the date seperator and that is for obvious reasons not a good character to have in a filename.

I have changed the formatter to use a fixed format so it is no longer locale dependant.

this will produce a file name like "Scan 2024-03-19 20.43.00.pdf"

### DIFF
--- a/swift-paperless/Views/DocumentScannerView.swift
+++ b/swift-paperless/Views/DocumentScannerView.swift
@@ -62,7 +62,11 @@ struct DocumentScannerView: UIViewControllerRepresentable {
                 }
             }
 
-            let date = Date().formatted(date: .numeric, time: .standard)
+            let date = Date().formatted(.verbatim(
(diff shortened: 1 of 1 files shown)


PTST:
As an aside i also changed the time seperator to a "."  
In many if not most locales the default would be ":"  
While that is allowed on unix systems including ios, this is only semi-supported on windows systems, so i think it is better to not use this.

paulgessinger:
@PTST Nice! I can't test this before the weekend but will merge asap once that's done.

roycepope:
Nice find! Looking forward to using this feature on iOS 😄

paulgessinger:
Ok, I managed to both reproduce the error now by switching to US locale, and then confirmed this PR fixes the error as expected. Merging now, thanks @PTST!

PTST:
Glad to hear it! 🎉

roycepope:
I pulled down build 42 and it works great!

Issue #70, opened 2024-04-01
Allow using installed client certificate


zindable:
In my configuration, I distinguish between external traffic originating from the internet and internal traffic within the local network. While internal traffic enjoys direct access to my services, external requests undergo additional authentication via mutual TLS (mTLS), terminated on my proxy.

I stumbled upon a [demo](https://github.com/MarcoEidinger/ClientCertificateSwiftDemo) showcasing how mTLS can be implemented in Swift. However, as I'm not proficient in Swift development, I'm uncertain about its feasibility.

Would it be possible for the app to show a list of installed profiles/certificates and utilize them in cases where the server mandates client authentication at the TLS layer?

paulgessinger:
I've received a few requests for this. I'm not an expert at all on the topic, so I'll have to research this first. It does seem useful though, so I'll look into it.

Nils-witt:
Hi, 

> Would it be possible for the app to show a list of installed profiles/certificates and utilize them in cases where the server mandates client authentication at the TLS layer?

@zindable As Apple limits the access to the OS Keychain that is not possible, but each app (or keychain group) can import them into their own keychain.

I´m currently trying to implement this feature here: [Fork](https://github.com/Nils-witt/swift-paperless/tree/mTLS) (WiP)

paulgessinger:
@Nils-witt awesome! Looking forward to a PR on this!

paulgessinger:
@zindable I pushed @Nils-witt's changes to TestFlight now. Could you give this a try?

Finkregh:
I tried to add a certificate, but got `certificateLoadError`. I used the same p12 file with https://github.com/immich-app/immich and it worked.
The log did not contain anything certificate related :/

paulgessinger:
@Finkregh I've found in my testing that it dependents on details of the certificate. 

I haven't tried this in Immich, but this app I using iOS API for this pretty much directly, so if it doesn't work it's highly likely it's because of Apple's framework. If you created the `.p12` with openssl3, I believe you need to pass `-legacy` to make it work. 

I found this out through trial and error trying to import a client cert first straight into macOS keychain and then into the app (the crypto framework is the same). 

Can you give that a shot?

Finkregh:
It works, nice!

In the Ui it's not clear how I can remove servers, Btw :)

paulgessinger:
[...]

PR #119, opened 2024-07-04
Feature: Support for mTLS

Nils-witt:
Added logic and UI for mTLS support to communicate with the api.
Issue: #70 

Added:
- UI for client certificate management.
- Logic for importing and storing the identities to the application keychain.
- Logic for assigning an Identity to a specific connection.
- Logic for using the identity in the ApiRepository and preview loaders.
- UI for mTLS authentification or only connection establishment.

Tested with:
- Paperless 2.10.2 
- Nginx (Proxy) 1.18.0 
- Ubuntu 22.04


paulgessinger:
Hey @Nils-witt! I'll try to go through this on the weekend. 

Do you have any pointers to how to set up mTLS with Paperless, so I can test everything?

Nils-witt:
There are different guides depending on the proxy you use.
I use Nginx and for testing i build myself this docker stack (Should be enough for testing): https://github.com/Nils-witt/paperless-nginx-mtls

It´s been a while since i initially configured Nginx for it, but this looks like it should work: https://medium.com/@mahernaija/how-to-configure-mutual-tls-mtls-for-nginx-736dec9f819d

paulgessinger:
@Nils-witt ah, so this is **on top** of whatever auth setup Paperless uses, just between the device and the reverse proxy?

Nils-witt:
Depending on your system there is the primary function to authorize the connection to the client from the proxy.
Additionally in some setups the certificate CN or similar is extracted and passed to paperless with the REMOTE_USER Header.

swift-paperless/Networking/PaperlessURLSessionDelegate.swift
```diff
@@ -0,0 +1,58 @@
...
+    public func loadIdentityByName(name: String?) {
+        credential = nil
+        if
+            let pName = name,
+            let identity = Keychain.readIdentity(name: pName)
+        {
+            credential = URLCredential(identity: identity, certificates: nil, persistence: .none)
+            return
+        } else {
+            Logger.shared.info("Error loading identity from keychain")
```
paulgessinger, line 23: Consider making this an error and throwing.

swift-paperless/Utilities/PKCS12.swift
```diff
@@ -0,0 +1,49 @@
...
+        let importPasswordOption: NSDictionary = [kSecImportExportPassphrase as NSString: password]
+        var items: CFArray?
+        let secError: OSStatus = SecPKCS12Import(pkcs12Data as NSData, importPasswordOption, &items)
+        guard secError == errSecSuccess else {
+            if secError == errSecAuthFailed {
+                throw PKCS12Error.wrongPassword
+            }
+            throw PKCS12Error.other
+        }
+        guard let theItemsCFArray = items else { fatalError() }
```
paulgessinger, line 28: Can this ever happen? Consider adding a case to `PKCS12Error for this.

swift-paperless/Utilities/PKCS12.swift
```diff
@@ -0,0 +1,49 @@
...
+        let theItemsNSArray: NSArray = theItemsCFArray as NSArray
+        guard let dictArray
+            = theItemsNSArray as? [[String: AnyObject]]
+        else {
+            fatalError()
```
paulgessinger, line 33: Same here.

swift-paperless/Utilities/PKCS12.swift
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        identity = dictArray.element(for: kSecImportItemIdentity)
+    }
+}
+
+extension [[String: AnyObject]] {
```
paulgessinger, line 40:
```suggestion
private extension [[String: AnyObject]] {
```
?

swift-paperless/Views/Settings/ConnectionsView.swift
```diff
@@ -62,7 +62,7 @@ struct ConnectionsView: View {
                 }
 
                 LabeledContent(String(localized: .settings(.activeServerUsername)), value: stored.user.username)
-
+                LabeledContent(String(localized: .settings(.activeIdentity)), value: stored.identity ?? "None")
```
paulgessinger, line 65: Can you localize `None` here?

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+enum CertificateState {
+    case notloaded
+    case wrongPassword
+    case valid
+}
+
+struct TLSListView: View {
+    @State var idenities: [TLSIdentity] = []
+    @Binding var identityNames: [String]
+    @State var showCreate: Bool = false
```
paulgessinger, line 27:
```suggestion
    @State var private idenities: [TLSIdentity] = []
    @Binding var identityNames: [String]
    @State private var showCreate: Bool = false
```

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+import SwiftUI
+
+struct TLSIdentity: Identifiable, Equatable {
+    var name: String
+    var identity: SecIdentity
+
+    var id: String { name }
+}
+
+enum CertificateState {
```
paulgessinger, line 18:
```suggestion
private enum CertificateState {
```

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+//  TLSListView.swift
+//  swift-paperless
+//
+//  Created by Nils Witt on 24.06.24.
+//
+
+import os
+import SwiftUI
+
+struct TLSIdentity: Identifiable, Equatable {
```
paulgessinger, line 11:
```suggestion
private struct TLSIdentity: Identifiable, Equatable {
```

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+}
+
+enum CertificateState {
+    case notloaded
+    case wrongPassword
+    case valid
+}
+
+struct TLSListView: View {
+    @State var idenities: [TLSIdentity] = []
```
paulgessinger, line 25:
```suggestion
    @State var identities: [TLSIdentity] = []
```

Requires some changes down in the file.

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+            idenities.removeAll()
+        }
+
+        identityNames.removeAll()
+        idenitites.forEach { ident, name in
+            withAnimation {
+                identityNames.append(name)
+                idenities.append(TLSIdentity(name: name, identity: ident))
+            }
+        }
```
paulgessinger, line 42:
```suggestion
        let identities: [(SecIdentity, String)] = Keychain.readAllIdenties()

        withAnimation {
            self.identities.removeAll()
        }

        identityNames.removeAll()
        identities.forEach { ident, name in
            withAnimation {
                identityNames.append(name)
                self.identities.append(TLSIdentity(name: name, identity: ident))
            }
        }
```

I would probably try to assemble the final value and then assign it in a single `withAnimation` call.

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+                    }, label: {
+                        Label(String(localized: .localizable(.add)), systemImage: "plus")
+                    })
+                    EditButton()
+                }
+            }
+        }
+        .sheet(isPresented: $showCreate, content: {
+            CreateView()
+                .navigationTitle("T")
```
paulgessinger, line 81: What does that mean?

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+                .toolbar {
+                    ToolbarItem(placement: .navigationBarTrailing) {
+                        HStack {
+                            Button {
+                                showCreate = true
+                                withAnimation {
+                                    // headers.append(.init(key: "Header", value: "Value"))
+                                }
```
paulgessinger, line 89: I'm guess this is a leftover.

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        private func saveToKeychain(certificateData: Data, certificatePassword: String, certificateName: String) {
+            do {
+                let pkc = try PKCS12(pkcs12Data: certificateData, password: certificatePassword)
+                if let identity = pkc.identity {
+                    Keychain.saveIdentity(identity: identity, name: certificateName)
+                }
+            } catch {
+                print(error)
```
paulgessinger, line 126: Use the logging for this. If it can realistically fail, it should probably be surface in the UI.

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+                    guard let selectedFile: URL = try result.get().first else { return }
+                    if selectedFile.startAccessingSecurityScopedResource() {
+                        certificateData = try Data(contentsOf: selectedFile)
+                        defer { selectedFile.stopAccessingSecurityScopedResource() }
+                        certificateName = selectedFile.lastPathComponent
+                    } else {
+                        Logger.shared.error("Error opening File from secure context")
+                    }
+                } catch {
+                    Logger.shared.error("erro while loding PfX \(error)")
```
paulgessinger, line 199:
```suggestion
                    Logger.shared.error("Error while loding PfX \(error)")
```

Consider surfacing these errors.

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+                }
+            }
+        }
+    }
+}
+
+struct TLSSingleView: View {
+    @State var identity: TLSIdentity
+    @State var cn: String?
```
paulgessinger, line 208:
```suggestion
private struct TLSSingleView: View {
    @State private var identity: TLSIdentity
    @State private var cn: String?
```

swift-paperless/Views/TLSListView.swift
```diff
@@ -0,0 +1,231 @@
...
+        .onChange(of: identity, initial: true) {
+            var optCertificate: SecCertificate?
+            SecIdentityCopyCertificate(identity.identity, &optCertificate)
+
+            if let certificate = optCertificate {
+                var cn: CFString?
+                SecCertificateCopyCommonName(certificate, &cn)
+
+                if cn != nil {
+                    self.cn = cn as String?
```
paulgessinger, line 226: Not super important, but I guess this could be a computed property of `TLSIdentity`.

swift-paperless/Views/Settings/SettingsView.swift
```diff
@@ -149,6 +149,11 @@ struct SettingsView: View {
             } label: {
                 Label(localized: .settings(.debugMenu), systemImage: "ladybug.fill")
             }
+            NavigationLink {
+                TLSListView(identityNames: Binding(get: { [] as [String] }, set: { _ in }))
```
paulgessinger, line 153: I'm not exactly sure what this does. Can this view actually modify the identities with this binding?

swift-paperless/Views/LoginView.swift
```diff
@@ -116,6 +122,14 @@ struct LoginView: View {
 
     @State private var loginOngoing = false
 
+    @State private var selectedIdentity: String? = nil
+
+    @State private var availableIdenityNames: [String] = []
```
paulgessinger, line 127:
```suggestion
    @State private var availableIdentityNames: [String] = []
```

swift-paperless/Networking/PaperlessURLSessionDelegate.swift
```diff
@@ -0,0 +1,58 @@
...
+//  PaperlessURLSessionDelegate.swift
+//  swift-paperless
+//
+//  Created by Nils Witt on 24.06.24.
+//
+
+import Foundation
+import os
+
+class PaperlessURLSessionDelegate: NSObject, URLSessionTaskDelegate {
```
paulgessinger, line 11:
```suggestion
final class PaperlessURLSessionDelegate: NSObject, URLSessionTaskDelegate {
```
This needs to be `final`, otherwise it can't conform to `Sendable` (which it must starting in Swift 6)

swift-paperless/Networking/PaperlessURLSessionDelegate.swift
```diff
@@ -0,0 +1,58 @@
...
+    private var credential: URLCredential? = nil
```
paulgessinger, line 12: This is a public mutable, and means the class becomes non-`Sendable`.
You could make this a `let` property, and switch from setting the identity via `loadIdentityByName` to supplying it to `init` instead.

paulgessinger (commented):
Overall, this looks really good!

Depending on how frequent errors are, it might make sense to include some error descriptions on the errors. The `ErrorController` has the ability to display extra info and also link out to documentation, which might be helpful here.

I haven't tested the functionality fully yet, but barring any major issues (which I don't expect), I'd be happy to put this in.

paulgessinger:
@Nils-witt but is the first use case supported? If I'm reading the code correctly, if you enter the mTLS path, the app does not store a token at all. How does that work then?

Nils-witt:
Thanks for the review
I will go over it later 

> @Nils-witt but is the first use case supported? If I'm reading the code correctly, if you enter the mTLS path, the app does not store a token at all. How does that work then?

This should be defined by the toggle (and the var identityBasedAuth) on the LoginView
If it´s not active it should follow the original flow and just add the delegate (changes to the func login, mostly l. 239ff.)

The func loginIentityBased is for the second use case.

swift-paperless/Utilities/PKCS12.swift
Nils-witt: This and the one below should never fail as the SecPKCS12Import function should not resolve to success when there were a problem and they should always be set that way. 
But for better handling I´ll add better error handling

swift-paperless/Views/Settings/SettingsView.swift
Nils-witt: Was a leftover from a quick fix

Nils-witt:
I´ve adapted your proposed changes.

paulgessinger (approved):
I haven't been able to test the actual certificate login myself, since I can't manage to add the certificate from your demo repository, but the changeset seems sensible.

I'll merge this now and push an update to TestFlight, in the hope that there are some folks on there that are able to give this a spin.

paulgessinger:
For future reference, it's necessary to use `openssl pkcs12 -legacy` when exporting a PKCS12 for iOS (or macOS, for that matter) (https://forums.developer.apple.com/forums/thread/697030)

PR #200, opened 2025-01-03
feat: Support limited permissions

paulgessinger:
Docs: https://docs.paperless-ngx.com/usage/#global-permissions

- [x] Test missing ui permissions setting + valid other settings
- [x] Refresh after permission removal should clear resources in store (see TODO in `DocumentStore.swift`).
- [x] Hide or grey-out the filter pills if we don't have view permissions
- [x] Communicate missing edit permissions in resource edit context
- [x] Prevent API error on saving new resource instances if we don't have create permissions
- [x] Communicate missing delete permissions in resource edit context
- [x] Investigate why it's possible to post documents even if no permissions
- [x] Check: Edit implies view (do I need to layer the tests client-side?)
- [x] Test document has [tags, correspondent] etc set that user does not have permissions for
- [x] Test document has owner set that user does not have permissions for


docs/common_issues/forbidden.md
```diff
@@ -4,6 +4,13 @@ title: "Error: Forbidden"
 
 # Error "No access: XXX"
 
+!!! note
+
+    In versions up to and including `v1.5.0`, the app required at least
+    read-access to most of the resources available in Paperless-ngx. Starting from
+    `v1.6.0`, only a minimal set of permissions is required to access the app. See
```
paulgessinger, line 11:
```suggestion
    `v1.8.0`, only a minimal set of permissions is required to access the app. See
```

PR #242, opened 2025-03-08
fix: Adjust to new document notes payload format

paulgessinger:
This changed in https://github.com/paperless-ngx/paperless-ngx/pull/8948

it was previously giving the full notes content inline with the document response like this:

```json
[
    {
        "id": 44,
        "note": "What?",
        "created": "2025-03-08T16:50:30.277365Z",
        "user": {
            "id": 2,
            "username": "user",
            "first_name": "",
            "last_name": ""
        }
    },
    {
        "id": 41,
        "note": "Note",
        "created": "2025-03-08T16:46:04.081142Z",
        "user": {
            "id": 2,
            "username": "user",
            "first_name": "",
            "last_name": ""
        }
    }
```

while now it only returns an array of the note IDs like this:

```json
[44, 45]
```

### DIFF
--- a/DataModel/Sources/DataModel/DocumentModel.swift
+++ b/DataModel/Sources/DataModel/DocumentModel.swift
@@ -17,6 +17,21 @@ public protocol DocumentProtocol: Codable {
     var storagePath: UInt? { get set }
 }
 
+public struct NotesPayload: Decodable, Equatable, Sendable, Hashable {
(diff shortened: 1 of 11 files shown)


shamoon:
https://github.com/paperless-ngx/paperless-ngx/pull/9336 should prevent the need for this. Thanks again

PR #461, opened 2026-02-03
add scanner options

smitstay:
**Summary**
  - Extracted scanner button patterns into ScannerSettingsConfiguration for centralized,
  version specific configuration
  - Added diagnostic logging that captures the full view hierarchy when scanner settings are
  applied

**Notes**
  - Scanner control matching currently uses English accessibility labels. On non-English iOS
  locales, the controls won't be found and settings will silently not apply. Diagnostic logs
  capture the localized labels, so adding support for other locales is straightforward as users
  report them.
  - The fallback configuration for unknown iOS versions reuses iOS 18 patterns. If they don't
  match, the scanner still works normally just without the custom settings.

**Test plan - verified on my device - iPhone 17 - iOS 26.2.1**
  - Verify flash and autoscan toggles still work on iOS 18 (English)
  - Verify scanner opens and functions normally when controls aren't found (e.g. non-English locale
   or unsupported iOS)
  - Check Console.app logs show view hierarchy diagnostics and setting results

### DIFF
--- a/swift-paperless/Localization/Settings.xcstrings
+++ b/swift-paperless/Localization/Settings.xcstrings
@@ -2402,6 +2402,39 @@
         }
       }
     },
+    "scannerAutoscanLabel" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Automatic Scanning"
+          }
+        }
+      }
+    },
+    "scannerFlashLabel" : {
+      "extractionState" : "manual",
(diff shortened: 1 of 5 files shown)


smitstay:
Sorry for the delay, it's been a busy week for me. This can definitely be feature gated. I can spend some time this week putting it behind a feature gate, writing tests, and fixing the lint issues. I just don't want to spend too much time not knowing if this wouldn't be considered. I'll update the PR. 

Do you do remote flagging or want it flagged at build time?

PR #506, opened 2026-04-09
fix: include leaf certificate in URLCredential for mTLS

IngmarStein:
Extract the leaf certificate from the SecIdentity and include it in the URLCredential's certificates array. Without this, URLSession may send an empty certificate list in the ClientHello during TLS client-auth, which some servers and proxies (e.g. Pomerium) may not handle gracefully.

I'm trying to protect Paperless-ngx and Home Assistant with mTLS because both apps support it. It works flawlessly for HA, but my identity-aware proxy (Pomerium) rejects SwiftPaperless because it can't find a certificate matching a given SAN. This is the only meaningful difference I could find between the two code bases and it looks plausible that if the matcher fails because of the empty certificate list.

### DIFF
--- a/Networking/Sources/Networking/Api/PaperlessURLSessionDelegate.swift
+++ b/Networking/Sources/Networking/Api/PaperlessURLSessionDelegate.swift
@@ -18,7 +18,7 @@ public final class PaperlessURLSessionDelegate: NSObject, URLSessionTaskDelegate
     {
       Logger.networking.info(
         "URLSessionDelegate initializing with identity: \(pName, privacy: .public)")
-      credential = URLCredential(identity: identity, certificates: nil, persistence: .none)
+      credential = Self.makeCredential(identity: identity)
     } else {
       Logger.networking.info("URLSessionDelegate initializing without identity")
       credential = nil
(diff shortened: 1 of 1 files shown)


paulgessinger:
Looks plausible. Do you have a way to test this?

IngmarStein:
I have no Apple Developer account, but I could test on Testflight.

paulgessinger:
I'm not generally opposed to this, but as per Apple's [documentation](https://developer.apple.com/documentation/foundation/urlcredential/init(identity:certificates:persistence:)), it seems like the `certificates:`	parameter is for *intermediate* certificates, not for the leaf, which should be contained in `identity:`.

How confident are you that this is not an issue with your certificate or your proxy configuration? (not trying to be rude, sorry, just want to gauge what the right thing to do here is). I'm slightly worried that including the leaf cert where the networking stack otherwise sends intermediate certs might cause issues in other people's configurations.

In the end there's not too much harm in shipping this in TestFlight, and reverting in case there are any bug reports, that's what betas are for in the end 😅.

IngmarStein:
I basically made this to match Home Assistant: https://github.com/home-assistant/iOS/blob/295a2a7c9fdcca479209723e1ad21778d302ff60/Sources/Shared/API/mTLS/ClientCertificate.swift#L184-L199
That app works with the same proxy configuration, hence I thought it was worth a try to match, but I see your point about this being for intermediate certificates. If this ends up in TestFlight and doesn't change the behavior, I'm happy to send another PR to revert :)

IngmarStein:
@paulgessinger this is the change adding this to HA: https://github.com/home-assistant/iOS/pull/4409

IngmarStein:
Is this part of 1.10.0 (184) or will it go into the next TestFlight build?

paulgessinger:
@IngmarStein it should be in 184 indeed. Do you see a change?

IngmarStein:
It is working now, but I also had to change my proxy config for a different reason, so I'm not 100% certain about the effectiveness. Sorry :/

PR #578, opened 2026-05-26
feat: On-disk document cache

paulgessinger:
---
This is **part 1 of 12 in a stack** made with GitButler:
- <kbd>&nbsp;12&nbsp;</kbd> #618 
- <kbd>&nbsp;11&nbsp;</kbd> #617 
- <kbd>&nbsp;10&nbsp;</kbd> #600 
- <kbd>&nbsp;9&nbsp;</kbd> #598 
- <kbd>&nbsp;8&nbsp;</kbd> #597 
- <kbd>&nbsp;7&nbsp;</kbd> #596 
- <kbd>&nbsp;6&nbsp;</kbd> #595 
- <kbd>&nbsp;5&nbsp;</kbd> #592 
- <kbd>&nbsp;4&nbsp;</kbd> #591 
- <kbd>&nbsp;3&nbsp;</kbd> #590 
- <kbd>&nbsp;2&nbsp;</kbd> #588 
- <kbd>&nbsp;1&nbsp;</kbd> #578 👈

### DIFF
--- a/Common/Sources/Common/ContentStore.swift
+++ b/Common/Sources/Common/ContentStore.swift
@@ -165,13 +165,12 @@ public struct ContentStore: Sendable {
     try writeSidecar(
       for: key, modified: modified, suggestedFilename: suggestedFilename)
 
-    let friendlyTarget = friendlyURL(
-      for: key, suggestedFilename: suggestedFilename)
-    if let oldCanonicalInode,
-      FileManager.default.fileExists(atPath: friendlyTarget.path),
-      inode(of: friendlyTarget) == oldCanonicalInode
-    {
-      try? FileManager.default.removeItem(at: friendlyTarget)
+    // Sweep any friendly link still pointing at the *previous* canonical
+    // inode — covers both (i) plain overwrite under the same name and
+    // (ii) a server-side rename where the old name's link would otherwise
+    // be orphaned forever.
+    if let oldCanonicalInode {
+      removeFriendlyLinks(matchingInode: oldCanonicalInode)
     }
 
     let friendly =
@@ -195,13 +194,31 @@ public struct ContentStore: Sendable {
 
   public func delete(_ key: Key) throws {
     let canonical = url(for: key)
-    if let sidecar = readSidecar(for: key) {
-      let friendly = friendlyURL(
-        for: key, suggestedFilename: sidecar.suggestedFilename)
-      try? FileManager.default.removeItem(at: friendly)
-    }
+    // Capture the inode before removing the blob so we can sweep every
+    // friendly link that points at it — including disambiguated names
+    // (`name (docID).ext`) that the sidecar's suggestedFilename alone
+    // cannot tell us about.
+    let canonicalInode = inode(of: canonical)
+
     try? FileManager.default.removeItem(at: canonical)
     try? FileManager.default.removeItem(at: sidecarURL(for: key))
+
+    if let canonicalInode {
+      removeFriendlyLinks(matchingInode: canonicalInode)
+    }
+  }
+
+  /// Removes every entry under `Friendly/` whose underlying inode equals
+  /// the given inode. Bounded by the size of `Friendly/`, which the OS
+  /// keeps small under storage pressure. Best-effort: failures are swallowed.
+  private func removeFriendlyLinks(matchingInode target: NSNumber) {
+    guard
+      let contents = try? FileManager.default.contentsOfDirectory(
+        at: friendlyRoot, includingPropertiesForKeys: nil)
+    else { return }
+    for entry in contents where inode(of: entry) == target {
+      try? FileManager.default.removeItem(at: entry)
+    }
   }
 
   // MARK: - Sidecar
--- a/Common/Sources/Common/ContentStore.swift
+++ b/Common/Sources/Common/ContentStore.swift
@@ -1,0 +1,331 @@
+//
+//  ContentStore.swift
+//  swift-paperless
+//
+//  Created by Paul Gessinger on 26.05.26.
+//
+
+import Foundation
+import os
+
+private let log = Logger(
+  subsystem: "com.paulgessinger.swift-paperless", category: "ContentStore")
+
+/// On-disk blob cache keyed by `(serverID, documentRemoteID, versionID, kind)`.
+///
+/// Lives in the app-group container so the Share Extension (and a future
+/// File Provider extension) can read it on a locked device. Files are written
+/// with `.completeUntilFirstUserAuthentication` protection on iOS; on macOS
+/// (host tests) the protection class is a no-op.
+///
+/// `Hit.friendlyURL` is a hardlink under `Caches/ContentStore/Friendly/`
+/// named after the server-provided filename, so consumers like
+/// `UIActivityViewController` show a recognizable name.
+public struct ContentStore: Sendable {
+  public enum Kind: String, Sendable, CaseIterable {
+    case original
+    case archive
+    case thumbnail
+
+    public var fileExtension: String {
+      switch self {
+      case .original, .archive: "pdf"
+      case .thumbnail: "bin"
+      }
+    }
+  }
+
+  public struct Key: Hashable, Sendable {
+    public let serverID: UUID
+    public let documentRemoteID: UInt
+    public let versionID: String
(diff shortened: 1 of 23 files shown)


paulgessinger:
@codex

Common/Sources/Common/ContentStore.swift
```diff
@@ -0,0 +1,196 @@
...
+    case original
+    case archive
+    case thumbnail
+
+    public var fileExtension: String {
+      switch self {
+      case .original, .archive: "pdf"
+      case .thumbnail: "bin"
+      }
+    }
```
paulgessinger, line 33: Will this cleanly extend to future file type support, where files can be non-pdfs?

Common/Sources/Common/ContentStore.swift
```diff
@@ -0,0 +1,196 @@
...
+      self.versionID = versionID
+      self.kind = kind
+    }
+  }
+
+  public enum StoreError: Error {
+    case appGroupUnavailable(identifier: String)
+  }
+
+  public static let appGroup = "group.com.paulgessinger.swift-paperless"
```
paulgessinger: Is this duplicated elsewhere? Centralize?

Common/Sources/Common/ContentStore.swift
```diff
@@ -0,0 +1,196 @@
...
+    }
+  }
+
+  public struct Key: Hashable, Sendable {
+    public let serverID: UUID
+    public let documentRemoteID: UInt
+    public let versionID: String
+    public let kind: Kind
+
+    public static let currentVersion = "current"
```
paulgessinger, line 42: What does this specific sentinal mean and how is it used?

Common/Sources/Common/ContentStore.swift
```diff
@@ -0,0 +1,196 @@
...
+
+    public init(
+      serverID: UUID, documentRemoteID: UInt,
+      versionID: String = Key.currentVersion, kind: Kind
+    ) {
+      self.serverID = serverID
+      self.documentRemoteID = documentRemoteID
+      self.versionID = versionID
```
paulgessinger, line 50: Version ID is the remote version number? New document versions register as separate document IDs, linked together. Is this accounted for?

Common/Sources/Common/URLSession+download.swift
```diff
@@ -0,0 +1,47 @@
...
+extension URLSession {
+  /// Streams the response of `request` to a temp file on disk and reports
+  /// progress via KVO on `URLSessionTask.progress`. Mirrors the shape of
+  /// `getData(for:progress:)` so callers can swap them with no API churn.
+  ///
+  /// The returned URL points into the system temp directory and is owned by
+  /// the caller — `URLSession` deletes it once the delegate method returns,
+  /// so the caller MUST move or replace it before this method returns.
+  /// Background `URLSession`s only support download/upload tasks (not `data`),
+  /// so this shape is what the Stage 14 background sync engine needs too.
```
paulgessinger: Don't mention the plans terminology in code comments. You can reference future plans in a generic way.

Networking/Sources/Networking/Api/ApiRepository.swift
```diff
@@ -457,32 +485,112 @@ extension ApiRepository: Repository {
 
   public func download(
     documentID: UInt, original: Bool = false, progress: (@Sendable (Double) -> Void)? = nil
```
paulgessinger, line 487: Do we still need this? Can we either deprecate or remove it?

Networking/Sources/Networking/Repository.swift
```diff
@@ -138,14 +146,26 @@ public protocol Repository: Sendable {
...
-  public func download(documentID: UInt, original: Bool) async throws -> URL? {
+  public func download(documentID: UInt, original: Bool) async throws -> URL {
     try await download(documentID: documentID, original: original, progress: nil)
   }
 
+  // Default impl for conformers that don't implement the Document-aware path:
+  // fall back to the id-keyed call. ApiRepository overrides this directly to
+  // avoid a redundant document fetch when threading staleness through the
+  // ContentStore.
+  public func download(
```
paulgessinger, line 161: Do we have such conformers? Is there a way to fix this architecturally?

Common/Sources/Common/ContentStore.swift
```diff
@@ -0,0 +1,196 @@
...
+
+  private struct Sidecar: Codable {
+    var modified: Date?
+    var writtenAt: Date
+  }
+
+  private func writeSidecar(for key: Key, modified: Date?) throws {
+    let sidecar = Sidecar(modified: modified, writtenAt: Date())
+    let encoder = JSONEncoder()
+    encoder.dateEncodingStrategy = .iso8601
```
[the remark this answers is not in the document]
paulgessinger: Good catch — fixed in `942a72d`. The sidecar now uses JSONEncoder/Decoder's default numeric date encoding instead of `.iso8601`, so the sub-second precision in paperless `modified` timestamps round-trips exactly and `read(_:freshAgainst:)` matches. `readReturnsURLWhenModifiedMatches` now uses a sub-second timestamp (`1234.567891`) as a regression guard — it fails under whole-second encoding.

Common/Sources/Common/URLSession+download.swift
paulgessinger: Done in `942a72d` — dropped the "Stage 14" reference; the comment now speaks generically about "a future background sync engine".

Common/Sources/Common/ContentStore.swift
paulgessinger: It was — also in `Common/UserDefaults.swift` (plus the `.entitlements` files, which have to keep the literal since Xcode reads them before any Swift runs). Centralized in `942a72d`: new `Common/AppGroup.swift` with `AppGroup.identifier`, referenced by both `ContentStore.appGroup` and `UserDefaults.group`. Left the `ConnectionManager` literals alone for now — that file is churned by 8 upstack commits, so it can adopt `AppGroup.identifier` post-stack without a mid-stack conflict.

Common/Sources/Common/ContentStore.swift
paulgessinger: `"current"` is a placeholder version id used while this stage keys blobs by `documentRemoteID` and has no real version-row id yet. It goes away in #592 (feat/document-versions), which replaces `versionID: String` + the `"current"` sentinel with a real `versionID: UInt` from the server. Tracked to address there rather than documenting a sentinel that's about to be deleted.

Common/Sources/Common/ContentStore.swift
paulgessinger: Not yet at this stage — here blobs are keyed by `documentRemoteID` only. Linked document versions are exactly what #592 (feat/document-versions) handles: the content-store Key becomes `(serverID, versionID: UInt, kind)` so each server-side version row addresses its own blob. Tracked to that PR.

Networking/Sources/Networking/Api/ApiRepository.swift
paulgessinger: It's removed upstack in #592 (feat/document-versions, commit 938ce421d) once every call site threads a full `Document` through for the staleness key. Kept here only so this stage compiles standalone; the deprecation/removal lands there. Tracked.

Networking/Sources/Networking/Repository.swift
paulgessinger: #592 reworks this: once `download(documentID:)` is gone, the "conformers that don't implement the Document-aware path" case disappears and this default collapses into a plain trampoline that just supplies the `original`/`progress` defaults. So the architectural cleanup you're pointing at happens there — tracked to that PR.

Common/Sources/Common/ContentStore.swift
paulgessinger: Mostly yes. The `fileExtension` here only names the blob's on-disk path inside the cache (`original.pdf`), it's never shown to the user — the share sheet takes its display name from `Document.originalFileName`/`archivedFileName` via `NSItemProvider.suggestedName`. So a non-PDF original would still be stored/served correctly; only the internal path suffix would read `.pdf`. When non-PDF originals are actually supported we'd derive the suffix from the server filename. Left as-is for now.

PR #591, opened 2026-05-28
refactor: Complete wire<->domain model split

paulgessinger:
This pull request removes the dependency on the `MetaCodable` library and replaces its usage with standard Swift `Codable` and property wrappers. It introduces new property wrappers for handling special encoding/decoding cases, updates model initializers for clarity, and refactors tests and models to use these new wrappers. The changes simplify the codebase and reduce external dependencies.

### Removal of MetaCodable

* Removed the `MetaCodable` package from both `Common` and `DataModel` targets and their dependencies (`Package.swift` files), and deleted all usages of its macros and helper coders throughout the codebase. [[1]](diffhunk://#diff-8ad2f438444870897bde9667870068a362675caa85996e3711c89c6d48331425L22) [[2]](diffhunk://#diff-8ad2f438444870897bde9667870068a362675caa85996e3711c89c6d48331425L36-R35) [[3]](diffhunk://#diff-8a48941184c285fb1d07198c104f68d18017d613dc48278a8672fcaa7575f591L20-R28) [[4]](diffhunk://#diff-c52fb6a272f76e080ec614b0eba75fb51aef2ac7b957cb8c0c747782f9f85d69L8-L10) [[5]](diffhunk://#diff-93ab74fd68f56aabf6f4aa69bbd4b07e202bb70d13da17a12489f1787ad4d602L9-R86) [[6]](diffhunk://#diff-474515b4032f071afc715e58a420e2eb57d34badaf8c317d78217a3ae30b39feL8-R10)

### New Codable Property Wrappers

* Added `@DateOnlyCodable` property wrapper to replace `DateOnlyCoder`, providing custom encoding/decoding for `Date` as `YYYY-MM-DD` strings.
* Added `@NullCodable` property wrapper to encode/decode optionals as explicit JSON `null`, replacing `NullCoder`.

### Model Refactoring

[...]

PR #590, opened 2026-05-28
refactor: Remove legacy document detail V3 and BezelKit

paulgessinger:
V4 is the only detail implementation now. Drops DocumentEditView,
the property-bar preference, and the documentDetailViewV4 feature flag.
Keeps DocumentDetailViewProtocol for a possible future alternate UI.

---
This is **part 3 of 12 in a stack** made with GitButler:
- <kbd>&nbsp;12&nbsp;</kbd> #618 
- <kbd>&nbsp;11&nbsp;</kbd> #617 
- <kbd>&nbsp;10&nbsp;</kbd> #600 
- <kbd>&nbsp;9&nbsp;</kbd> #598 
- <kbd>&nbsp;8&nbsp;</kbd> #597 
- <kbd>&nbsp;7&nbsp;</kbd> #596 
- <kbd>&nbsp;6&nbsp;</kbd> #595 
- <kbd>&nbsp;5&nbsp;</kbd> #592 
- <kbd>&nbsp;4&nbsp;</kbd> #591 
- <kbd>&nbsp;3&nbsp;</kbd> #590 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #588 
- <kbd>&nbsp;1&nbsp;</kbd> #578

PR #588, opened 2026-05-28
feat: Needs-auth and offline state

paulgessinger:
---
This is **part 2 of 12 in a stack** made with GitButler:
- <kbd>&nbsp;12&nbsp;</kbd> #618 
- <kbd>&nbsp;11&nbsp;</kbd> #617 
- <kbd>&nbsp;10&nbsp;</kbd> #600 
- <kbd>&nbsp;9&nbsp;</kbd> #598 
- <kbd>&nbsp;8&nbsp;</kbd> #597 
- <kbd>&nbsp;7&nbsp;</kbd> #596 
- <kbd>&nbsp;6&nbsp;</kbd> #595 
- <kbd>&nbsp;5&nbsp;</kbd> #592 
- <kbd>&nbsp;4&nbsp;</kbd> #591 
- <kbd>&nbsp;3&nbsp;</kbd> #590 
- <kbd>&nbsp;2&nbsp;</kbd> #588 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #578

PR #592, opened 2026-05-28
feat: Document versions model + version-keyed content store

paulgessinger:
This pull request introduces support for document versioning in the data model, networking layer, and content cache. The main changes add a new `DocumentVersion` type, update the `Document` model to include a list of versions, and ensure that downloads and thumbnails are version-aware. The content cache keying is revised to use version IDs, and the API layer is updated to parse and use version information from the backend. Tests are added and updated to validate the new versioning behavior.

**Data Model and Versioning Support:**
* Added a new `DocumentVersion` struct and a `versions: [DocumentVersion]` property to the `Document` model, along with computed properties `currentVersionID` and `rootVersionID` for resolving the correct version ID. [[1]](diffhunk://#diff-474515b4032f071afc715e58a420e2eb57d34badaf8c317d78217a3ae30b39feR55-R73) [[2]](diffhunk://#diff-474515b4032f071afc715e58a420e2eb57d34badaf8c317d78217a3ae30b39feR100-R106) [[3]](diffhunk://#diff-474515b4032f071afc715e58a420e2eb57d34badaf8c317d78217a3ae30b39feR180-R194)
* Updated the model initializer and equality/hash implementations to include `versions`. [[1]](diffhunk://#diff-474515b4032f071afc715e58a420e2eb57d34badaf8c317d78217a3ae30b39feR138) [[2]](diffhunk://#diff-474515b4032f071afc715e58a420e2eb57d34badaf8c317d78217a3ae30b39feR159)
* Added comprehensive tests for version resolution logic in `DocumentVersionTest.swift`.

**Networking Layer Updates:**
* Updated API wire types to parse and expose the `versions` array from the backend, including a new `ApiDocumentVersion` struct and mapping logic. [[1]](diffhunk://#diff-72306145cdc8cd791186ae863aa57ba1168e28ad2942e5a716528b19eb15bd60R40-R41) [[2]](diffhunk://#diff-72306145cdc8cd791186ae863aa57ba1168e28ad2942e5a716528b19eb15bd60R65) [[3]](diffhunk://#diff-4da7ac3c8b5eb599e396e16e9ad68697bde09cf7746ff6ce1957881ae882345fR1-R31)
* Refactored repository interfaces and implementations to require a `Document` (not just an ID) for downloads, ensuring version and modified date are available for correct cache validation. [[1]](diffhunk://#diff-f0918b1347afcd63fa90f692953627c496a8187f2a76e55a8f662ecd2c9465f5L464-R498) [[2]](diffhunk://#diff-f0918b1347afcd63fa90f692953627c496a8187f2a76e55a8f662ecd2c9465f5L534-R529) [[3]](diffhunk://#diff-495b980c6a2a194333e3ee8002de5910907515fc1f81046e3ba53b793dd303beL281-R282) [[4]](diffhunk://#diff-a4b007092547c91479c5cb5e697eb66b0d0cb9dd8dac28dd659b3e8de32f55d9L497-R498) [[5]](diffhunk://#diff-a4b007092547c91479c5cb5e697eb66b0d0cb9dd8dac28dd659b3e8de32f55d9L507-R512) [[6]](diffhunk://#diff-11c730b4da38686f182277ede7813f26d61bc15d02c44df3abefaeaacac404c8L25-R26)
* Updated endpoints for downloads and thumbnails to accept an optional version parameter and include it in API requests. [[1]](diffhunk://#diff-e6b8c18691e50bd6b3dc5751ba3ed21e029a2cde5ffdb6850ba997c98f0bc290L146-R163) [[2]](diffhunk://#diff-f0918b1347afcd63fa90f692953627c496a8187f2a76e55a8f662ecd2c9465f5L520-R511) [[3]](diffhunk://#diff-f0918b1347afcd63fa90f692953627c496a8187f2a76e55a8f662ecd2c9465f5L827-R819)

**Content Cache (ContentStore) Changes:**
* Changed `ContentStore.Key` to use only `(serverID, versionID, kind)` (removing `documentRemoteID` and string version IDs), and updated all usages accordingly. [[1]](diffhunk://#diff-b11372b13579e5990e8ca49436f04ad6c63d0513314fd3943d162ac22c381de9L11-R18) [[2]](diffhunk://#diff-b11372b13579e5990e8ca49436f04ad6c63d0513314fd3943d162ac22c381de9L38-L49) [[3]](diffhunk://#diff-b11372b13579e5990e8ca49436f04ad6c63d0513314fd3943d162ac22c381de9L89-R89) [[4]](diffhunk://#diff-8414b4ede75152478f37da91f3176c55cd5ea64776f8cdc4c202cde0984f0b07L35-R57)
* Added tests to ensure the cache path includes the version ID.

**Repository Interface and Implementation Cleanup:**
* Removed old download methods that accepted only document IDs, enforcing the use of the new version-aware methods.

These changes ensure that all document downloads, thumbnails, and cached content are correctly tied to the appropriate document version, providing robust support for multi-version documents and future-proofing for backend enhancements.

## Behavioural notes (raised in review)

Three consequences of this change that are easy to miss from the diff. All of them are
bounded by the same fact: **multi-version support exists only from paperless-ngx
v3.0.0-beta.rc1**, so on every 2.x backend `versions` is empty, `currentVersionID` is
just `id`, and behaviour is unchanged from before this PR.

**1. A stale `Document` now selects a version, not merely a stale copy.**
`currentVersionID` is read off the caller's `Document`, and two production call sites can
hold one that predates a server-side version bump — `DocumentPreview.loadDocument` (from
the list row) and `DocumentDetailModel`, which deliberately fires a provisional download
before the metadata round-trip completes. Previously that staleness only cost a cache
check; now it can also pick a different file. The provisional download is superseded once
`store.document(id:)` returns, so the effect is a briefly-rendered older version rather
than a stuck one — but it is the biggest behavioural shift here.

**2. `?version=` is only sent when the server reported versions.** `currentVersionID`
falls back to `document.id`, so an unconditional query parameter would go out on *every*
download and thumbnail request. Nuke keys its data cache on the URL string, so that would
invalidate every cached thumbnail for every user on first launch after upgrade — to
convey an id the backend would have picked anyway. `Document.versionQueryID` is `nil`
when `versions` is empty, keeping URLs byte-identical for non-versioned servers. This
also means the parameter never reaches a backend that predates the feature.

**3. Version-keyed `ContentStore` paths remove the bound on cache growth.** Under the old
key a re-download overwrote the same path, so the store was self-limiting at one blob per
document per kind. Keyed by version, each new version gets its own directory and nothing
reclaims the old one — and there is no eviction today (`ContentStore.delete(_:)` has no
production callers; `purge()` is reachable only from the debug "clear local storage"
action). Stage 14's refcounting is the intended fix, so this ships a real steady-state
disk regression that persists for several PRs. Unbounded on a versioned backend, a no-op
on any 2.x one.

### Verification

Version-resolution behaviour is covered by unit tests only — **this has not been
exercised against a live multi-version backend.** The default-version ordering was
matched against the paperless-ngx source (`get_latest_version_for_root` uses
`order_by("-id")`, `documents/versioning.py`), not observed empirically.

**Stack** (bottom to top):

<details>

- ~~#592 feat/document-versions~~ (merged)  👈

</details>

- #618 feat/multi-server-sync
- #627 feat/background-sync

### DIFF
--- a/Common/Sources/Common/ContentStore.swift
+++ b/Common/Sources/Common/ContentStore.swift
@@ -8,7 +8,14 @@
 import Foundation
 import os
 
-/// On-disk blob cache keyed by `(serverID, documentRemoteID, versionID, kind)`.
+/// On-disk blob cache keyed by `(serverID, versionID, kind)`.
+///
+/// `versionID` is the server-side document version row id; since paperless-ngx
+/// stores versions as sibling rows in the document table, version ids share a
+/// namespace with document ids, so a single id uniquely identifies the
+/// content. For documents without a `versions` array (older backends, or
+/// single-file docs) callers pass the document id directly — it equals the
+/// root version id server-side.
 ///
 /// Lives in the app-group container so the Share Extension (and a future
 /// File Provider extension) can read it on a locked device. Files are written
@@ -35,18 +42,11 @@ public struct ContentStore: Sendable {
 
   public struct Key: Hashable, Sendable {
     public let serverID: UUID
-    public let documentRemoteID: UInt
-    public let versionID: String
+    public let versionID: UInt
     public let kind: Kind
 
-    public static let currentVersion = "current"
-
-    public init(
-      serverID: UUID, documentRemoteID: UInt,
-      versionID: String = Key.currentVersion, kind: Kind
-    ) {
+    public init(serverID: UUID, versionID: UInt, kind: Kind) {
       self.serverID = serverID
-      self.documentRemoteID = documentRemoteID
       self.versionID = versionID
       self.kind = kind
     }
@@ -86,8 +86,7 @@ public struct ContentStore: Sendable {
   private func directory(for key: Key) -> URL {
     canonicalRoot
       .appendingPathComponent(key.serverID.uuidString, isDirectory: true)
-      .appendingPathComponent(String(key.documentRemoteID), isDirectory: true)
-      .appendingPathComponent(key.versionID, isDirectory: true)
+      .appendingPathComponent(String(key.versionID), isDirectory: true)
   }
 
   public func url(for key: Key) -> URL {
--- a/DataModel/Sources/DataModel/Model/DocumentModel.swift
+++ b/DataModel/Sources/DataModel/Model/DocumentModel.swift
@@ -52,6 +52,25 @@ public struct DocumentNote: Identifiable, Equatable, Sendable, Hashable {
   }
 }
 
+public struct DocumentVersion: Identifiable, Equatable, Sendable, Hashable {
+  public var id: UInt
+  public var added: Date
+  public var label: String?
+  public var checksum: String?
+  public var isRoot: Bool
+
+  public init(
+    id: UInt, added: Date, label: String? = nil,
+    checksum: String? = nil, isRoot: Bool
+  ) {
+    self.id = id
+    self.added = added
+    self.label = label
+    self.checksum = checksum
+    self.isRoot = isRoot
+  }
+}
+
 public struct Document: Identifiable, Equatable, Hashable, Sendable {
   public var id: UInt
   public var title: String
@@ -77,6 +96,13 @@ public struct Document: Identifiable, Equatable, Hashable, Sendable {
 
   public var notes: NotesPayload = .init()
   public var customFields: CustomFieldRawEntryList
+
+  // Server-side version rows for this document, newest-first or order-undefined
+  // depending on backend. Empty on backends predating multi-version support.
+  // `currentVersionID` and `rootVersionID` resolve the right id even when this
+  // is empty by falling back to `self.id` — the document id equals the root
+  // version id on the server (versions are sibling rows in the same table).
+  public var versions: [DocumentVersion] = []
 
   // Presense of this depends on the endpoint
   // If we didn't get a value, we likely just modified
@@ -109,6 +135,7 @@ public struct Document: Identifiable, Equatable, Hashable, Sendable {
     pageCount: Int? = nil,
     notes: NotesPayload = .init(),
     customFields: CustomFieldRawEntryList = CustomFieldRawEntryList(),
+    versions: [DocumentVersion] = [],
     userCanChange: Bool = true,
     permissions: Permissions? = nil,
     setPermissions: Permissions? = nil
@@ -129,6 +156,7 @@ public struct Document: Identifiable, Equatable, Hashable, Sendable {
     self.pageCount = pageCount
     self.notes = notes
     self.customFields = customFields
+    self.versions = versions
     self.userCanChange = userCanChange
     self.permissions = permissions
     self.setPermissions = setPermissions
@@ -148,6 +176,21 @@ extension Document {
     if let name = serverName, !name.isEmpty { return name }
     let base = title.isEmpty ? "document" : title
     return "\(base).pdf"
+  }
+
+  /// Id of the version the server returns when no `?version=` query is sent.
+  /// paperless serves the newest version by default; we mirror that by picking
+  /// the latest `added`. Falls back to `self.id` when `versions` is empty
+  /// (older backends, single-file documents) — the document id equals the
+  /// root version id server-side, so this fallback is correct, not a fudge.
+  public var currentVersionID: UInt {
+    versions.max(by: { $0.added < $1.added })?.id ?? id
+  }
+
+  /// Id of the original/root version. Same fallback rationale as
+  /// `currentVersionID`.
+  public var rootVersionID: UInt {
+    versions.first(where: \.isRoot)?.id ?? id
   }
 }
 
--- a/DataModel/Sources/DataModel/Model/DocumentModel.swift
+++ b/DataModel/Sources/DataModel/Model/DocumentModel.swift
@@ -97,8 +97,10 @@ public struct Document: Identifiable, Equatable, Hashable, Sendable {
   public var notes: NotesPayload = .init()
   public var customFields: CustomFieldRawEntryList
 
-  // Server-side version rows for this document, newest-first or order-undefined
-  // depending on backend. Empty on backends predating multi-version support.
+  // Server-side version rows for this document. The serializer sorts these
+  // id-descending (newest first), but nothing here relies on that — the
+  // accessors below order explicitly. Empty on backends predating
+  // multi-version support, which means anything before v3.0.0-beta.
   // `currentVersionID` and `rootVersionID` resolve the right id even when this
   // is empty by falling back to `self.id` — the document id equals the root
   // version id on the server (versions are sibling rows in the same table).
@@ -179,12 +181,29 @@ extension Document {
   }
 
   /// Id of the version the server returns when no `?version=` query is sent.
-  /// paperless serves the newest version by default; we mirror that by picking
-  /// the latest `added`. Falls back to `self.id` when `versions` is empty
-  /// (older backends, single-file documents) — the document id equals the
-  /// root version id server-side, so this fallback is correct, not a fudge.
+  /// paperless resolves that with `order_by("-id").first()`, so we order by
+  /// id too — *not* by `added`, which can disagree after an import or restore
+  /// preserves original timestamps. Ids are unique primary keys, so this
+  /// ordering is total and the result never depends on array order.
+  ///
+  /// Falls back to `self.id` when `versions` is empty (older backends,
+  /// single-file documents) — the document id equals the root version id
+  /// server-side, so this fallback is correct, not a fudge.
   public var currentVersionID: UInt {
-    versions.max(by: { $0.added < $1.added })?.id ?? id
+    versions.max(by: { $0.id < $1.id })?.id ?? id
+  }
+
+  /// Version id to put in a `?version=` query, or `nil` when the parameter
+  /// should be omitted entirely.
+  ///
+  /// With no versions, `currentVersionID` is `self.id` — the root id, which is
+  /// exactly what the backend picks anyway, so sending it conveys nothing.
+  /// Omitting it keeps request URLs byte-identical to what pre-versions builds
+  /// sent, which matters because Nuke keys its thumbnail data cache on the URL
+  /// string: sending a redundant parameter would invalidate every cached
+  /// thumbnail for every user whose server has no multi-version support.
+  public var versionQueryID: UInt? {
+    versions.isEmpty ? nil : currentVersionID
   }
 
   /// Id of the original/root version. Same fallback rationale as
--- a/Networking/Sources/Networking/Api/ApiRepository.swift
+++ b/Networking/Sources/Networking/Api/ApiRepository.swift
@@ -462,49 +462,40 @@ extension ApiRepository: Repository {
   }
 
   public func download(
-    documentID: UInt, original: Bool = false, progress: (@Sendable (Double) -> Void)? = nil
-  ) async throws -> URL {
-    // No Document handle → no modified timestamp to validate the cache,
-    // so streamDownload's nil-modified guard bypasses ContentStore and
-    // fetches fresh every call. Callers should prefer download(document:...).
-    try await streamDownload(
-      documentID: documentID, original: original,
-      modified: nil, progress: progress)
-  }
-
-  public func download(
     document: Document, original: Bool = false,
     progress: (@Sendable (Double) -> Void)? = nil
   ) async throws -> URL {
-    try await streamDownload(
-      documentID: document.id, original: original,
-      modified: document.modified, progress: progress)
+    try await streamDownload(document: document, original: original, progress: progress)
   }
 
   private func streamDownload(
-    documentID: UInt, original: Bool,
-    modified: Date?, progress: (@Sendable (Double) -> Void)?
+    document: Document, original: Bool,
+    progress: (@Sendable (Double) -> Void)?
   ) async throws -> URL {
     Logger.networking.notice("Downloading document (original: \(original))")
+
+    let version = document.currentVersionID
 
     // Without a staleness signal (modified timestamp) we can't validate a
     // cached blob — and writing one back without `modified` would cache it
     // indefinitely with no way to detect server-side changes. Bypass the
     // ContentStore entirely in that case. Also fall back when the app-group
     // container isn't available (host tests, mis-configured entitlement).
-    guard let contentStore, let modified else {
+    guard let contentStore, let modified = document.modified else {
       return try await fetchToTemp(
-        documentID: documentID, original: original, progress: progress)
+        documentID: document.id, original: original, version: version,
+        progress: progress)
     }
 
     let key = ContentStore.Key(
       serverID: connection.serverID,
-      documentRemoteID: documentID,
+      versionID: version,
       kind: original ? .original : .archive)
 
     if let cached = contentStore.read(key, freshAgainst: modified) {
       Logger.networking.info(
-        "ContentStore hit for documentID \(documentID) (original: \(original))")
+        "ContentStore hit for documentID \(document.id) version \(version) (original: \(original))"
+      )
       progress?(1.0)
       return cached
     }
@@ -517,7 +508,7 @@ extension ApiRepository: Repository {
       defer { inFlightDownloads[key] = nil }
 
       let request = try request(
-        .download(documentId: documentID, original: original))
+        .download(documentId: document.id, original: original, version: version))
       let (tempURL, response) = try await urlSession.getDownload(
         for: request, progress: progress)
 
@@ -531,11 +522,11 @@ extension ApiRepository: Repository {
   }
 
   private func fetchToTemp(
-    documentID: UInt, original: Bool,
+    documentID: UInt, original: Bool, version: UInt?,
     progress: (@Sendable (Double) -> Void)?
   ) async throws -> URL {
     let request = try request(
-      .download(documentId: documentID, original: original))
+      .download(documentId: documentID, original: original, version: version))
     let (tempURL, response) = try await urlSession.getDownload(
       for: request, progress: progress)
     try validateDownloadResponse(response, request: request)
@@ -824,7 +815,8 @@ extension ApiRepository: Repository {
     func thumbnailRequest(document: Document) throws -> URLRequest
   {
     Logger.networking.debug("Get thumbnail for document \(document.id, privacy: .public)")
-    let url = try url(Endpoint.thumbnail(documentId: document.id))
+    let url = try url(
+      Endpoint.thumbnail(documentId: document.id, version: document.currentVersionID))
 
     var request = URLRequest(url: url)
     addTokenTo(request: &request)
--- a/Networking/Sources/Networking/Api/ApiRepository.swift
+++ b/Networking/Sources/Networking/Api/ApiRepository.swift
@@ -36,7 +36,19 @@ public class ApiRepository {
 
   // Per-key in-flight task map: two concurrent downloads of the same blob
   // share one network request rather than racing into the ContentStore.
-  private var inFlightDownloads: [ContentStore.Key: Task<URL, Error>] = [:]
+  //
+  // Keyed by the staleness stamp as well as the content key. Coalescing on the
+  // content key alone would merge callers holding the same version but
+  // different `modified` timestamps, and whichever won the race would write its
+  // own stamp into the sidecar — recording the blob as fresher than it is if
+  // the loser held the newer stamp, which `read(_:freshAgainst:)` would then
+  // happily serve.
+  private struct DownloadKey: Hashable {
+    let content: ContentStore.Key
+    let modified: Date
+  }
+
+  private var inFlightDownloads: [DownloadKey: Task<URL, Error>] = [:]
 
   nonisolated
     private let apiVersion: UInt?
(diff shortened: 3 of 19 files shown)


paulgessinger:
This PR is part of a stack:

1. [`refactor/wire-domain-split`](https://github.com/paulgessinger/swift-paperless/pull/591)
1. **`feat/document-versions` <-- this PR**
1. [`feat/persistence-foundation`](https://github.com/paulgessinger/swift-paperless/pull/595)
1. [`refactor/observable-document-store`](https://github.com/paulgessinger/swift-paperless/pull/596)
1. [`feat/element-cache-records`](https://github.com/paulgessinger/swift-paperless/pull/597)
1. [`feat/element-cache-source-of-truth`](https://github.com/paulgessinger/swift-paperless/pull/598)
1. [`feat/document-cache-source-of-truth`](https://github.com/paulgessinger/swift-paperless/pull/600)
1. [`feat/offline-entire-library`](https://github.com/paulgessinger/swift-paperless/pull/617)
1. [`feat/multi-server-sync`](https://github.com/paulgessinger/swift-paperless/pull/618)
1. [`fix/pdf-preview-cache-latency`](https://github.com/paulgessinger/swift-paperless/pull/620)
1. [`feat/background-sync`](https://github.com/paulgessinger/swift-paperless/pull/627)

<details><summary>3 earlier closed/merged PRs</summary>

1. ~~[`feat/needs-auth-and-offline`](https://github.com/paulgessinger/swift-paperless/pull/588)~~
1. ~~[`refactor/remove-detail-v3`](https://github.com/paulgessinger/swift-paperless/pull/590)~~
1. ~~[`feat/element-cache-aside`](https://github.com/paulgessinger/swift-paperless/pull/619)~~

</details>

---
*Created with [jjpr](https://github.com/michaeldhopkins/jjpr)*

paulgessinger:
@codex

Networking/Sources/Networking/Repository.swift
```diff
@@ -155,24 +150,14 @@ public protocol Repository<Documents, Tasks>: Sendable {
...
+  // Trampoline that supplies defaults for callers that don't need a progress
+  // callback or always want the archive variant. Delegates straight to the
+  // protocol requirement.
   public func download(
     document: Document, original: Bool = false,
     progress: (@Sendable (Double) -> Void)? = nil
   ) async throws -> URL {
-    try await download(
-      documentID: document.id, original: original, progress: progress)
+    try await download(document: document, original: original, progress: progress)
```
paulgessinger: **Verified — this trampoline is the protocol's default witness, and it calls itself.**

This extension method has the same signature as the protocol requirement (default argument values don't participate in witness matching), so it *is* the default implementation — and its body dispatches to the witness, which is itself. Any conformer that doesn't implement `download(document:original:progress:)` gets infinite recursion, and **compiles without a warning**.

Confirmed with a minimal repro — a conformer that omits the method type-checks cleanly:

```swift
protocol R {
  func download(document: Int, original: Bool, progress: (@Sendable (Double) -> Void)?) async throws -> String
}
extension R {
  func download(document: Int, original: Bool = false, progress: (@Sendable (Double) -> Void)? = nil) async throws -> String {
    try await download(document: document, original: original, progress: progress)
  }
}
struct Conformer: R {}   // compiles — the extension satisfies the requirement
```

Before this PR the extension body called a *different* requirement (`download(documentID:)`), so it was a genuine default impl. Deleting the bare-id path is what made it self-referential.

No live bug — all five conformers implement it today. But the compiler no longer enforces that, and the failure mode is a hang rather than an error. Cheapest fix: give the trampoline a distinct signature so it can't satisfy the requirement, or drop it and put the defaults on the requirement itself.

Networking/Sources/Networking/Api/ApiRepository.swift
```diff
@@ -461,50 +461,41 @@ extension ApiRepository: Repository {
...
 
   private func streamDownload(
-    documentID: UInt, original: Bool,
-    modified: Date?, progress: (@Sendable (Double) -> Void)?
+    document: Document, original: Bool,
+    progress: (@Sendable (Double) -> Void)?
   ) async throws -> URL {
     Logger.networking.notice("Downloading document (original: \(original))")
 
+    let version = document.currentVersionID
```
paulgessinger, line 493: **Question — a stale `Document` now selects the wrong version, not just a stale one.**

`currentVersionID` is read off the caller's `Document`. Two of the three production call sites pass a document that may predate a server-side version bump:

- `DocumentPreview.loadDocument` takes its `Document` straight from the list row.
- `DocumentDetailModel` (~L202) deliberately fires a *provisional* download from the pre-refresh document before the metadata round-trip completes.

Under the old key that only cost a stale-cache check. Now the same staleness also picks the version id, i.e. potentially a different file rather than an older copy of the same one.

Is the re-validation pass after `store.document(id:)` guaranteed to correct the provisional download, or can a user sit on the previous version until the list refreshes? Worth a line in the PR description either way — it's the biggest behavioural shift here and it isn't called out.

Networking/Sources/Networking/Api/ApiRepository.swift
```diff
@@ -824,7 +815,8 @@ extension ApiRepository: Repository {
     func thumbnailRequest(document: Document) throws -> URLRequest
   {
     Logger.networking.debug("Get thumbnail for document \(document.id, privacy: .public)")
-    let url = try url(Endpoint.thumbnail(documentId: document.id))
+    let url = try url(
+      Endpoint.thumbnail(documentId: document.id, version: document.currentVersionID))
```
paulgessinger: **Noted — every thumbnail URL changes for every user on upgrade, not just versioned documents.**

Because `currentVersionID` falls back to `document.id`, this is never nil, so `?version=` is appended to **every** thumbnail request — including against backends with no multi-version support.

Nuke keys its data cache on the URL string (`ImagePipeline.Cache.makeDataCacheKey` → `request.preferredImageId` → `urlRequest.url?.absoluteString`), which is exactly the mechanism the commit message relies on for free per-version partitioning. The flip side is that it invalidates the entire existing thumbnail cache for the whole user base on first launch after upgrade.

The migration note only discusses `ContentStore` blobs, so this cost looks unaccounted for. It's transient and probably acceptable — but it's a full thumbnail re-fetch, which is visible on a slow connection. Worth stating explicitly rather than discovering later.

Networking/Sources/Networking/Api/Endpoint.swift
```diff
@@ -143,15 +143,24 @@ extension Endpoint {
...
-  public static func download(documentId: UInt, original: Bool = false) -> Endpoint {
+  public static func download(
+    documentId: UInt, original: Bool = false, version: UInt? = nil
+  ) -> Endpoint {
     var queryItems: [URLQueryItem] = []
     if original {
       queryItems.append(URLQueryItem(name: "original", value: "true"))
     }
+    if let version {
+      queryItems.append(URLQueryItem(name: "version", value: String(version)))
```
paulgessinger, line 162: **Question — has the `?version=` query been exercised against a pre-versions server?**

Following from the thumbnail comment: `version` is optional here, but every production call site passes a non-nil `currentVersionID`, so this parameter now goes out on effectively every download and thumbnail request regardless of backend version.

DRF generally ignores unrecognized query parameters, so this is very likely fine — but "very likely" is carrying real weight in a sentence about every file request the app makes, given the app supports a range of server versions. If it's been tested against an older instance, worth noting; if not, worth doing before this lands.

Common/Sources/Common/ContentStore.swift
```diff
@@ -86,8 +86,7 @@ public struct ContentStore: Sendable {
   private func directory(for key: Key) -> URL {
     canonicalRoot
       .appendingPathComponent(key.serverID.uuidString, isDirectory: true)
-      .appendingPathComponent(String(key.documentRemoteID), isDirectory: true)
-      .appendingPathComponent(key.versionID, isDirectory: true)
+      .appendingPathComponent(String(key.versionID), isDirectory: true)
```
paulgessinger, line 89: **Noted — version-keyed paths remove the natural bound on cache growth.**

This goes a bit further than the orphaned-blob point in the commit message. Under the old key, re-downloading a document *overwrote* the same path, so the store was self-bounding at one blob per document per kind. Keyed by version, each new version gets its own directory and nothing reclaims the old one.

There's no eviction path today: `ContentStore.delete(_:)` has no production callers, and `purge()` is reachable only from `DocumentStore.wipeLocalCache()` (the debug "clear local storage" action). No size cap, no sweep.

Probably fine to defer given Stage 14 is explicitly refcounting — but it's a real regression in steady-state disk behaviour that ships now and gets fixed several PRs later, so I'd rather it were acknowledged in the description than silent.

Networking/Sources/Networking/Api/ApiRepository.swift
```diff
@@ -461,50 +461,41 @@ extension ApiRepository: Repository {
...
       return try await fetchToTemp(
-        documentID: documentID, original: original, progress: progress)
+        documentID: document.id, original: original, version: version,
+        progress: progress)
     }
 
     let key = ContentStore.Key(
       serverID: connection.serverID,
-      documentRemoteID: documentID,
+      versionID: version,
```
paulgessinger, line 509: **Question — in-flight coalescing now merges callers with different freshness.**

`inFlightDownloads` is keyed by `ContentStore.Key`, which after this change contains nothing caller-specific. Two callers holding the same version but *different* `modified` timestamps now share a single task, and whichever won the race writes its `modified` into the sidecar.

If the loser's timestamp was the newer one, the cached blob ends up recorded as fresher than it actually is, and `read(_:freshAgainst:)` will serve it. Narrow window, and it may well be unreachable in practice — but the key collapse is what widened it, so it seems worth a thought here rather than later.

(Separately, and pre-existing: a caller joining an in-flight task gets no `progress` callbacks, since only the first caller's closure is captured.)

DataModel/Sources/DataModel/Model/DocumentModel.swift
```diff
@@ -149,6 +177,21 @@ extension Document {
...
     return "\(base).pdf"
   }
+
+  /// Id of the version the server returns when no `?version=` query is sent.
+  /// paperless serves the newest version by default; we mirror that by picking
+  /// the latest `added`. Falls back to `self.id` when `versions` is empty
+  /// (older backends, single-file documents) — the document id equals the
+  /// root version id server-side, so this fallback is correct, not a fudge.
+  public var currentVersionID: UInt {
+    versions.max(by: { $0.added < $1.added })?.id ?? id
```
paulgessinger: **Minor — ties in `added` resolve arbitrarily.**

`max(by:)` returns an unspecified element when two versions share an `added` timestamp. Almost certainly unreachable in practice.

Worth mentioning only because `DocumentVersionTest.versionOrderIndependent` explicitly asserts that array order doesn't affect resolution — and in the tie case that property quietly doesn't hold. A secondary sort on `id` would make the guarantee real for free:

```swift
versions.max { ($0.added, $0.id) < ($1.added, $1.id) }?.id ?? id
```

paulgessinger (commented):
_These review comments were generated by an AI agent (Claude Code) and posted by @paulgessinger. Only the finding marked **verified** was reproduced (with a compiler check); the rest are readings of the code that warrant a human check._

---

Review pass over the versioning model and the content-store re-key. The shape of the change makes sense to me — collapsing the key now rather than retrofitting Stages 8/13/14 around a `"current"` sentinel is the right call, and deleting the bare-id download path outright (rather than leaving a cache-bypassing overload reachable) is the right instinct.

Seven comments below, ordered by what I'd address first. They differ a lot in confidence, so each says which it is:

- **1 verified** — the `Repository` trampoline is the protocol's default witness and calls itself. Reproduced with a compiler check. No live bug today, but the compiler stopped enforcing what used to be enforced.
- **3 questions** — stale `Document` now selecting a version rather than just a staleness check; `?version=` against pre-versions backends; in-flight coalescing merging callers with different `modified`.
- **3 noted** — full thumbnail-cache invalidation on upgrade, unbounded cache growth, and an `added` tie-break that undercuts one test's stated property.

Only the first was reproduced; the rest are readings of the code that you're better placed to settle than I am.

DataModel/Sources/DataModel/Model/DocumentModel.swift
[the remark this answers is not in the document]
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `0621782c` — you were right, and I confirmed it against the backend source rather than reasoning about it.

`get_latest_version_for_root` resolves the default version with `manager.filter(root_document=root_doc).order_by("-id").first()` ([`src/documents/versioning.py:48-55`](https://github.com/paperless-ngx/paperless-ngx/blob/dev/src/documents/versioning.py)), so it is strictly highest-id — not `added`, and not `version_index` (which only feeds filename suffixes in `file_handling.py:135-136`). `currentVersionID` now orders by `id`.

This also subsumes the tie-break point raised separately: ids are unique primary keys, so the ordering is total and no longer depends on array order.

Two tests added, because none of the existing ones discriminated — every fixture had `added` and `id` agreeing:
- highest id wins even when its `added` is *older* (the import/restore case you described)
- equal `added` timestamps still resolve deterministically

paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `0621782c`, though not with the secondary-sort spelling.

Ordering by `id` alone (see the sibling thread — the backend uses `order_by("-id")`, so `added` was the wrong key to begin with) makes the tie case vanish rather than resolving it: ids are unique primary keys, so the ordering is already total.

You were right that `versionOrderIndependent` was asserting a property that did not hold. Added `equalAddedTimestampsResolveByID` to cover the case explicitly, since that test would have kept passing either way.

Networking/Sources/Networking/Repository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `0621782c`. Confirmed the diagnosis — and it is a nasty one precisely because the compiler goes quiet.

Fixed by narrowing the trampoline to `download(document:original:)`, which cannot satisfy the three-argument requirement, so the compiler resumes enforcing that every conformer implements it. Kept the comment explaining *why* the signature is deliberately short, since the obvious "tidy-up" here is to add `progress:` back and silently reintroduce the recursion.

Checked the call sites first: only one goes through the protocol (`DocumentDetailModel.swift:133`, two-arg), and the tests call the concrete `ApiRepository` method, whose own defaults still apply. Nothing needed changing.

Networking/Sources/Networking/Api/ApiRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `0621782c` — fixed rather than just documented, because the cost turned out to be worse than "transient".

Multi-version support only exists from **v3.0.0-beta.rc1** (`git tag --contains` on the feature commit lists no 2.x release), so `versions` is empty for effectively the entire current user base. Sending `?version=` unconditionally would therefore have flushed every thumbnail for nearly every user in order to convey a version id the backend would have picked anyway.

Added `Document.versionQueryID`, which is `nil` when `versions.isEmpty`. Thumbnail and download URLs are now byte-identical to what pre-versions builds sent whenever the server has no versions, so Nuke's data cache survives the upgrade untouched.

The `ContentStore` key still uses the non-optional `currentVersionID` — that is a local path, not a URL, and it should stay version-partitioned. The two are now named distinctly in `streamDownload` with a comment, since conflating them is exactly the trap here.

Networking/Sources/Networking/Api/Endpoint.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Answered and made moot in `0621782c`.

**Answer:** it is safe on old backends. The parameter is read with a bare `request.query_params.get("version")` ([`versioning.py:29-32`](https://github.com/paperless-ngx/paperless-ngx/blob/dev/src/documents/versioning.py)) — no serializer, no `OpenApiParameter` declaration anywhere in `views.py` — so on any backend predating the feature it is simply an unrecognized query param and DRF ignores it. Your instinct was right; I just did not want it resting on "very likely" either.

**Made moot:** per the thumbnail thread, `?version=` is no longer sent at all unless the server actually reported versions, so it never reaches a pre-versions backend in the first place.

For the record, on a backend that *does* support it: an unparseable value → 404 `"Invalid version parameter"`; a valid integer that is not part of that document's version family → plain 404. Both are avoided by only ever sending ids the server itself gave us.

Still not exercised against a live pre-v3 instance — this is a source read, not a run.

Networking/Sources/Networking/Api/ApiRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `0621782c` for the half this PR introduced.

`inFlightDownloads` is now keyed by `DownloadKey { content, modified }` rather than the bare `ContentStore.Key`, so two callers holding the same version but different staleness stamps no longer share a task and cannot record the loser's timestamp in the sidecar. Added `concurrentDownloadsWithDifferentModifiedNotDeduped` to pin it.

Narrow as you say, but the fix is five lines and the failure mode (a blob recorded as fresher than it is, then served by `read(_:freshAgainst:)`) is silent, so it did not seem worth leaving.

**Your parenthetical is not fixed** — a caller joining an in-flight task still gets no `progress` callbacks, since only the first caller's closure is captured. It is genuinely pre-existing, so I left it out of a review-fix commit, but I have tracked it: two views opening the same document at once leave one with a dead progress bar.

Networking/Sources/Networking/Api/ApiRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

No code change — answering, and adding it to the PR description as you asked.

**To the direct question:** no, the re-validation pass is not a *guarantee*, but the staleness is transient rather than sticky. The provisional download is superseded once `store.document(id:)` returns and the view re-renders off the refreshed document; a user cannot sit on the previous version until a list refresh, because the detail load itself performs that round-trip. What they can get is the older version briefly rendered first.

**Why it is not being fixed here:** the exposure requires a server-side version bump landing between the list fetch and the open, *and* a backend that supports versions at all — which means v3.0.0-beta.rc1 or later. On every 2.x backend `versions` is empty, `currentVersionID` is just `id`, and the behaviour is byte-for-byte what it was before this PR. So the regression window is real but currently unreachable for essentially all users.

You are right that it is the biggest behavioural shift in the change and was not called out. Added to the PR description, along with the honest caveat that none of this has been exercised against a live multi-version backend — the version behaviour is covered by unit tests only.

Common/Sources/Common/ContentStore.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Agreed on the substance, holding on the fix — and acknowledged in the description rather than left silent, which is what you asked for.

Your framing is more accurate than the commit message's: this is not just orphaned blobs, it removes the *bound*. The old key made the store self-limiting at one blob per document per kind because a re-download overwrote in place; version-keyed paths have no such property, and I confirmed there is no eviction path today — `ContentStore.delete(_:)` has no production callers and `purge()` is reachable only from the debug "clear local storage" action.

Held rather than fixed because Stage 14 is explicitly refcounting, and a stopgap sweep here would be work that stage then has to unpick. Tracked with a specific note that Stage 14 must reclaim *superseded versions*, not merely deleted documents — the distinction that makes this a regression rather than a pre-existing gap.

Worth being clear about what that means: this ships a real steady-state disk regression that stays unfixed for several PRs. It is bounded in practice by the same v3.0.0-beta.rc1 constraint as the rest of the versioning behaviour — no versions, no extra directories — but on a versioned backend it is unbounded until Stage 14.

paulgessinger:

Worked through all 8 review comments. **5 applied** in `0621782c` (one follow-up commit appended to this PR's range), **1 held** for Stage 14, **2 answered** without code changes. Every thread has an individual reply.

Before triaging I read the paperless-ngx source at `v3.0.0-beta.rc2` rather than reasoning about backend behaviour, which settled two of the comments outright:

- **Default version resolution is `order_by("-id").first()`** (`documents/versioning.py:48-55`) — so `currentVersionID` ordering by `added` was simply wrong, and the codex comment was correct. Ordering by id is also total (ids are unique PKs), which subsumes the separate tie-break finding.
- **`?version=` is read with a bare `request.query_params.get(...)`** (`versioning.py:29-32`), no serializer and no OpenAPI declaration, so old backends silently ignore it. Safe — but now moot, since it is no longer sent to them.
- **Multi-version support exists only from v3.0.0-beta.rc1** — no 2.x release has it. This bounds every behavioural concern in the review and is now stated in the description.

The two changes worth a second look, since neither was literally what the comments asked for:

1. **`Document.versionQueryID`** — `nil` when `versions` is empty, so `?version=` is omitted entirely rather than sent redundantly. The comment only asked for the thumbnail-cache cost to be acknowledged; fixing it seemed better than documenting it, given it would have flushed thumbnails for essentially the whole user base. The `ContentStore` key deliberately still uses the non-optional `currentVersionID` — local path, not URL.
2. **The `Repository` trampoline** is now `download(document:original:)`, deliberately without `progress:`, so it cannot become the requirement's own default witness. Re-adding the third parameter would silently reintroduce the infinite recursion, which is why the comment there is longer than the code.

**Held:** unbounded cache growth from version-keyed paths (Stage 14 refcounting). This ships a genuine steady-state disk regression that stays unfixed for several PRs; it is now called out in the description rather than left silent, and tracked with the note that Stage 14 must reclaim *superseded versions*, not just deleted documents.

**Not fixed, pre-existing:** a caller joining an in-flight download still receives no `progress` callbacks — only the first caller's closure is captured.

### Verification

- `swift test` green on DataModel (213) and Networking (203), including 5 new tests. Worth noting that **none of the pre-existing version tests discriminated the id-vs-`added` bug** — every fixture had the two agreeing.
- `build_sim` green at this PR's head and at the stack tip.
- Rebasing 72 descendants produced 9 conflicted commits, all cascades of a single hunk (an upstack commit wraps the same call in `RequestError.normalizing`); resolved in the conflicted commit itself. Stack is clean and fully pushed.
- **Not exercised against a live multi-version backend** — all version behaviour is unit-tested only, and the ordering fix is matched against backend source rather than observed. That is the main gap if you want a manual pass before this lands.

PR #596, opened 2026-05-31
refactor: Migrate DocumentStore and ConnectionManager to @Observable

paulgessinger:
**Stack** (bottom to top):

<details>

- ~~#596 refactor/observable-document-store~~ (merged)  👈

</details>

- #618 feat/multi-server-sync
- #627 feat/background-sync

### DIFF
--- a/AppShared/Sources/AppShared/Networking/ConnectionManager.swift
+++ b/AppShared/Sources/AppShared/Networking/ConnectionManager.swift
@@ -6,7 +6,6 @@
 //
 
 import AuthenticationServices
-import Combine
 import Common
 import CryptoKit
 import DataModel
@@ -165,7 +164,8 @@ public struct StoredConnection: Equatable, Identifiable, Sendable {
 }
 
 @MainActor
-public class ConnectionManager: ObservableObject {
+@Observable
+public class ConnectionManager {
   private struct PreviewLaunchArguments {
     let mode: Bool?
     let url: String?
@@ -223,19 +223,21 @@ public class ConnectionManager: ObservableObject {
     }
   }
 
-  public enum Event {
+  public enum Event: Sendable {
     case connectionChange(animated: Bool)
     case logout
   }
 
-  public var eventPublisher =
-    PassthroughSubject<Event, Never>()
+  public let events = Broadcaster<Event>()
 
+  @ObservationIgnored
   private let previewArguments: PreviewLaunchArguments?
   public let previewMode: Bool
 
+  @ObservationIgnored
   private let database: Database
-  private var observationTask: Task<Void, Never>?
+  @ObservationIgnored
+  private nonisolated(unsafe) var observationTask: Task<Void, Never>?
 
   public init(database: Database, previewMode: Bool? = nil) {
     self.database = database
@@ -294,11 +296,7 @@ public class ConnectionManager: ObservableObject {
   /// that follows it — is authoritative and reconciles this dict against the
   /// table; being equality-guarded, it is a no-op when the eager write already
   /// matched.
-  public private(set) var connections: [UUID: StoredConnection] = [:] {
-    willSet {
-      objectWillChange.send()
-    }
-  }
+  public private(set) var connections: [UUID: StoredConnection] = [:]
 
   /// Active server pointer.
   ///
@@ -306,18 +304,32 @@ public class ConnectionManager: ObservableObject {
   /// Share Extension picks up active-server changes through UserDefaults' free
   /// cross-process syncing. The "dangling pointer after row delete" case is
   /// handled in ``applyHydrate(records:)`` and ``logout(animated:)``.
+  ///
+  /// The @UserDefaultsBacked property wrapper is incompatible with the
+  /// @Observable macro's auto-tracking, so observation is wired manually
+  /// via access(keyPath:) / withMutation(keyPath:) around the private
+  /// backing storage.
+  @ObservationIgnored
   @UserDefaultsBacked("ActiveConnectionId", storage: .group)
-  public var activeConnectionId: UUID? = nil {
-    willSet {
-      objectWillChange.send()
+  private var _activeConnectionId: UUID? = nil
+
+  public var activeConnectionId: UUID? {
+    get {
+      access(keyPath: \.activeConnectionId)
+      return _activeConnectionId
+    }
+    set {
+      withMutation(keyPath: \.activeConnectionId) {
+        _activeConnectionId = newValue
+      }
     }
   }
 
   /// Per-connection "needs auth" set. Hydrated from the `needs_auth` column;
-  /// kept as a published Set<UUID> here so the existing banner / lock-badge
+  /// kept as a Set<UUID> here so the existing banner / lock-badge
   /// `.onChange(of: needsAuthIds)` watchers keep firing without touching the
   /// SwiftUI sites in this commit.
-  @Published public private(set) var needsAuthIds: Set<UUID> = []
+  public private(set) var needsAuthIds: Set<UUID> = []
 
   /// Apply a fresh snapshot of records to the in-memory dict and the
   /// needs-auth set. Called by both the bootstrap read and the observer.
@@ -380,7 +392,7 @@ public class ConnectionManager: ObservableObject {
   // the app shell observes this and presents `ReauthSheet`. Decoupled from
   // `needsAuthIds` so the banner can show without auto-presenting a sheet —
   // user consent stays explicit.
-  @Published public var reauthRequested: UUID? = nil
+  public var reauthRequested: UUID? = nil
 
   public func requestReauth(for id: UUID) {
     Logger.api.info(
@@ -394,7 +406,7 @@ public class ConnectionManager: ObservableObject {
 
   public func setActiveConnection(id: UUID, animated: Bool = true) {
     activeConnectionId = id
-    eventPublisher.send(.connectionChange(animated: animated))
+    events.emit(.connectionChange(animated: animated))
   }
 
   public func isServerUnique(_ url: URL) -> Bool {
@@ -539,11 +551,11 @@ public class ConnectionManager: ObservableObject {
       } else {
         Logger.api.info("Setting active connection to nil")
         self.activeConnectionId = nil
-        eventPublisher.send(.logout)
+        events.emit(.logout)
       }
     } else {
       activeConnectionId = nil
-      eventPublisher.send(.logout)
+      events.emit(.logout)
     }
   }
 
--- a/AppShared/Sources/AppShared/Networking/ConnectionManager.swift
+++ b/AppShared/Sources/AppShared/Networking/ConnectionManager.swift
@@ -165,7 +165,7 @@ public struct StoredConnection: Equatable, Identifiable, Sendable {
 
 @MainActor
 @Observable
-public class ConnectionManager {
+public final class ConnectionManager {
   private struct PreviewLaunchArguments {
     let mode: Bool?
     let url: String?
@@ -237,7 +237,7 @@ public class ConnectionManager {
   @ObservationIgnored
   private let database: Database
   @ObservationIgnored
-  private nonisolated(unsafe) var observationTask: Task<Void, Never>?
+  private var observationTask: Task<Void, Never>?
 
   public init(database: Database, previewMode: Bool? = nil) {
     self.database = database
--- a/AppShared/Sources/AppShared/Networking/ConnectionManager.swift
+++ b/AppShared/Sources/AppShared/Networking/ConnectionManager.swift
@@ -305,10 +305,17 @@ public final class ConnectionManager {
   /// cross-process syncing. The "dangling pointer after row delete" case is
   /// handled in ``applyHydrate(records:)`` and ``logout(animated:)``.
   ///
-  /// The @UserDefaultsBacked property wrapper is incompatible with the
-  /// @Observable macro's auto-tracking, so observation is wired manually
-  /// via access(keyPath:) / withMutation(keyPath:) around the private
-  /// backing storage.
+  /// A property wrapper cannot sit on an `@Observable`-tracked property at all:
+  /// the macro rewrites `x` into computed accessors over generated storage named
+  /// `_x`, which collides with the wrapper's own `_x` and fails to compile with
+  /// *"ambiguous reference to member '_x'"*. So the wrapper keeps the storage,
+  /// `@ObservationIgnored` keeps the macro off it, and the public property is
+  /// hand-written using the same `access` / `withMutation` pair the macro would
+  /// have generated.
+  ///
+  /// Note the key path is the **public** ``activeConnectionId``, not the private
+  /// storage: the registrar has to be keyed on what callers actually read, or
+  /// observers would never be notified.
   @ObservationIgnored
   @UserDefaultsBacked("ActiveConnectionId", storage: .group)
   private var _activeConnectionId: UUID? = nil
--- a/AppShared/Sources/AppShared/Utilities/View+onEvent.swift
+++ b/AppShared/Sources/AppShared/Utilities/View+onEvent.swift
@@ -1,0 +1,65 @@
+import Persistence
+import SwiftUI
+
+extension View {
+  /// Listen to a ``Broadcaster`` for the lifetime of this view's `@State`
+  /// storage. Shape matches `.onReceive(publisher:perform:)` but is backed by
+  /// an `AsyncStream` subscription whose `Task` outlives view
+  /// disappear/reappear cycles (NavigationStack push/pop) — `.task` and
+  /// `for await` would cancel on disappear and drop events.
+  public func onEvent<Element>(
+    from broadcaster: Broadcaster<Element>,
+    perform action: @escaping @MainActor (Element) -> Void
+  ) -> some View {
+    modifier(BroadcasterEventModifier(broadcaster: broadcaster, action: action))
+  }
+}
+
+/// Indirection so the long-lived subscription always invokes the handler from
+/// the *current* view value.
+///
+/// The subscription is created once and deliberately outlives disappear /
+/// reappear, so it cannot capture the handler directly: a closure captured at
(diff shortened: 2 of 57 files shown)


paulgessinger (commented):
> **This review was written by Claude (Claude Code), acting on my behalf.** Notes on provenance are at the bottom: what was verified by reading the tree at `8851df8d9` vs. what is inferred.

**Overall:** clean, mechanical, and correct. `DocumentStore`/`ConnectionManager` → `@Observable`, `PassthroughSubject` → `Broadcaster`, ~50 call sites converted `@EnvironmentObject` → `@Environment(T.self)`. Nothing here looks broken. Four things worth addressing, none blocking.

Stack note: #595 is merged, so this PR now sits directly on `main` and its diff is exactly its own three commits.

## Things this quietly fixes

`DocumentCell`, `DocumentList`, `CommonPickerFilterView` etc. take the store as a **plain `var store: DocumentStore`**, not `@ObservedObject`. Under `ObservableObject` those views were never invalidated by store changes directly — they only refreshed because some ancestor holding `@EnvironmentObject` re-rendered the whole subtree. Under `@Observable` the reads are genuinely tracked. Net: fewer redraws *and* fewer latent staleness bugs.

The two behaviour changes the commits assert both check out:

- Dropping Combine's `removeDuplicates()` in `observeFriendlyName` is safe — `setFriendlyName` early-returns on an equal normalized value (`ConnectionManager.swift:518`).
- The hand-rolled `withObservationTracking` loop is correct despite `onChange` being willSet-timed: the task re-reads `store.settings.appTitle` *after* resuming on the main actor, so it always observes the post-mutation value.

## 1. `BroadcasterEventModifier` diverges from the `.onReceive` semantics it claims

`Subscription.swift:96-102` subscribes once in `onAppear` and never again:

```swift
content.onAppear {
  if subscription == nil { subscription = broadcaster.sink(action) }
}
```

Two consequences the doc comment ("Shape matches `.onReceive(publisher:perform:)`") doesn't cover:

- **`action` is frozen at first appear.** `.onReceive`'s `SubscriptionView` swaps in the latest closure on every update; this captures the view struct once. Any plain `let`/`var` view property read inside the handler is permanently stale.
- **A new `broadcaster` is ignored.** If the identity changes while the view stays alive, events keep flowing from the dead one.

All four current call sites are safe: the handlers only touch `@State`-boxed view models, `FilterModel`/`DocumentDetailModel` reference types, and `@EnvironmentObject` singletons, and `store` is created exactly once (`swift_paperlessApp.swift:171-173` — the `else` branch only runs when `store == nil`), so `store.events` never changes identity.

But this is *public API in a shared package* and the trap is invisible at the call site. Routing the action through a `@State`-held reference box updated in `body`, plus an `.onChange(of: ObjectIdentifier(broadcaster))` resubscribe, would make it actually match the documented shape.

## 2. `friendlyNameTask` no longer self-cancels

`swift_paperlessApp.swift:28` swapped `@State var friendlyNameSubscription: AnyCancellable?` for `@State var friendlyNameTask: Task<Void, Never>?`. `AnyCancellable` cancelled itself when `@State` storage was freed; a bare `Task` does not — it runs forever, retaining `store` and `manager`.

`UIApplicationSupportsMultipleScenes` is `false` and `observeFriendlyName` is called at most once, so `MainView` lives for the process lifetime and **this leaks nothing today**. It's hygiene — but note that commit 3 adds `Subscription`, a handle that cancels on `deinit`, for exactly this shape, and this is the one site that would benefit and doesn't use it.

## 3. `nonisolated(unsafe)` on the two task properties

`DocumentStore.swift:69` and `ConnectionManager.swift:240` both gained `nonisolated(unsafe)` alongside `@ObservationIgnored`. Both are mutated from the main actor and read from a nonisolated `deinit` — safe in practice (deinit implies no other references), but the annotation also disables checking for any *future* nonisolated access, and neither commit message explains why it became necessary.

Worth confirming it was genuinely required by the `@Observable` expansion rather than added to clear an unrelated diagnostic. If it was required, an `isolated deinit` or a small `Sendable` box is the narrower tool.

## 4. Smaller notes

- **Layering:** the SwiftUI `View.onEvent` extension lives in `Persistence`, so `DocumentList.swift` now does `import Persistence` purely to get a view modifier. `Broadcaster` belonging there is defensible; the UI glue less so — `Common` or `AppShared` fits better.
- **Dead API:** `sink(_:)`, `store(in: inout Set<Subscription>)`, and the `RangeReplaceableCollection` overload are unused — and still unused eleven PRs up the stack, where `onEvent` remains the only consumer (4 sites). Either use them for #2 above, or trim to what `onEvent` needs.
- `ConnectionManager` stayed non-`final` while gaining `@Observable`; making it `final` avoids the subclass-observation footgun.
- `Subscription` can be plain `Sendable` — `Task` is already `Sendable`, so `@unchecked` buys nothing.
- The title overstates scope: `ErrorController`, `AppSettings`, `BiometricLockManager`, `DocumentImportModel`, `AttachmentManager` and two others are still `ObservableObject`. Fine for a stacked PR, just not what the title says.

## What's worth smoke-testing

The fine-grained-invalidation switch is the main risk surface and CI doesn't cover it: edit a document from the list (swipe to strip inbox tags), rename a tag/correspondent and confirm cells update, and trigger the task-error toast. Those paths previously rode on blanket `objectWillChange` fan-out.

---

### Provenance

Written by Claude Code from a read of the tree at `8851df8d9` plus the three commit diffs. CI was green on all jobs at review time.

**Verified by reading code:** `setFriendlyName`'s idempotency guard; that `store` is constructed exactly once in `refreshConnection`; that all four `onEvent` handlers only mutate reference- or `@State`-backed storage; that `sink`/`store(in:)` have no callers anywhere in the stack up to `feat/background-sync`; `UIApplicationSupportsMultipleScenes = false` in `Info.plist`; the set of types still conforming to `ObservableObject`.

**Inferred, not verified:** the `.onReceive`/`SubscriptionView` action-refresh semantics in point 1 come from SwiftUI's documented behaviour, not from testing this build. Point 3 is a question, not a finding — the reason `nonisolated(unsafe)` became necessary was not determined. Nothing was built or run locally; no runtime behaviour was exercised.

paulgessinger:

Worked through the review. Seven of the eight items are applied in `0ab159ac`, appended at the end of this PR's range; item 2 landed on **#627** instead, for a reason worth recording. Build green at this PR's head and at the stack tip, zero warnings; 116 Persistence tests pass.

### Two corrections to the review

**`sink(_:)` is not dead code.** It is the only mechanism behind `BroadcasterEventModifier` (`subscription = broadcaster.sink(action)`). Genuinely unused were the two `store(in:)` overloads — plus the `Hashable` conformance that existed solely to make `Set<Subscription>` possible. Those are gone; `sink` stays.

**Item 3 was a finding, not just a question.** Four minimal repros — `@MainActor` + `@Observable` + `@ObservationIgnored` + a nonisolated `deinit` reading the task, with and without explicit `Sendable`, final and non-final — all typecheck under `-swift-version 6` *without* `nonisolated(unsafe)`. `@ObservationIgnored` leaves the property a plain stored property, so the same rule that let the pre-migration code compile still applies. Both annotations removed, app build green.

### Items

1. **`BroadcasterEventModifier` semantics** — fixed. The action is now refreshed on every `body` evaluation through a `@State`-held box, and `.onChange(of: ObjectIdentifier(broadcaster))` resubscribes when the identity changes. The review's read of the call sites checked out — all four only touch `@State`-boxed or reference storage — so this is public-API robustness rather than a live bug. The box carries `@unchecked Sendable`, and there it is load-bearing (written in `body`, read in the sink handler, both main-actor) unlike the one removed from `Subscription`.

2. **`friendlyNameTask` self-cancellation** — applied on **#627** as `03aa6d9f`, not here. Applying it here conflicted **66 commits**: seven upstack commits across #597/#617/#618/#627 edit that exact declaration line, and `swift_paperlessApp.swift` was the only conflicted path in the whole stack. Since the stack merges as a unit the end state is identical, and the leak is theoretical anyway — `UIApplicationSupportsMultipleScenes` is `false` and `observeFriendlyName` runs at most once, so `MainView` lives for the process lifetime. Not worth hand-resolving seven commits and re-verifying 66 rebased ones for that.

3. **`nonisolated(unsafe)`** — removed from both properties. See above.

4. **Smaller notes**
   - *Layering:* `View.onEvent` + the modifier moved to `AppShared/Utilities/View+onEvent.swift`. `Broadcaster` stays in `Persistence` — it does back GRDB table observation. Your read was exact: `DocumentList.swift` imported `Persistence` for *nothing but* this modifier, even at the stack tip, so that import came out. `CreateDocumentView`, `DocumentDetailViewV4` and `swift_paperlessApp` keep theirs — they use `Database`.
   - *Dead API:* trimmed as described above.
   - *`final ConnectionManager`:* done — no subclasses anywhere in the stack.
   - *`Subscription: Sendable`:* done, `@unchecked` dropped. `init` is now `public`, which #627's use needs.
   - *Title:* retitled to name the two types actually migrated.

### Still not exercised at runtime

Your smoke-test list stands and is the right list — fine-grained invalidation is precisely what CI cannot cover. Item 1 adds one case: a NavigationStack push/pop with an event arriving while a child view is on screen, which is the scenario the modifier exists for.

---

**Provenance:** verified by building — app at this PR's head and at the stack tip (`xcodegen generate` at each), `swift test --package-path Persistence` (116 tests), and the four standalone repros for item 3. Verified by reading: that `store(in:)` and `Hashable` have no callers or test references anywhere in the stack; that `DocumentList` uses no other `Persistence` symbol even at the tip; that `ConnectionManager` has no subclasses; which seven upstack commits touch the `friendlyName` declaration. **Not verified:** no runtime behaviour was exercised — nothing was driven in the simulator, so the invalidation paths above remain untested. The `.onReceive`/`SubscriptionView` action-refresh semantics behind item 1 still come from SwiftUI's documented behaviour, not from observing this build.

paulgessinger:
@codex

AppShared/Sources/AppShared/Utilities/View+onEvent.swift
```diff
@@ -0,0 +1,65 @@
...
+    return
+      content
+      .onAppear {
+        if subscription == nil {
+          subscription = makeSubscription()
+        }
+      }
+      // A broadcaster swapped underneath a live view would otherwise keep
+      // delivering from the old one.
+      .onChange(of: ObjectIdentifier(broadcaster)) { _, _ in
```
paulgessinger, line 71: What does this pattern do?

AppShared/Sources/AppShared/Utilities/View+onEvent.swift
```diff
@@ -0,0 +1,65 @@
...
+/// handler, both of which are main-actor isolated.
+private final class EventActionBox<Element>: @unchecked Sendable {
+  var action: (@MainActor (Element) -> Void)?
+}
+
+private struct BroadcasterEventModifier<Element: Sendable>: ViewModifier {
+  let broadcaster: Broadcaster<Element>
+  let action: @MainActor (Element) -> Void
+
+  @State private var box = EventActionBox<Element>()
```
paulgessinger, line 50: What's `EventActionBox` and what does it do?

AppShared/Sources/AppShared/Networking/ConnectionManager.swift
```diff
@@ -294,30 +296,40 @@ public class ConnectionManager: ObservableObject {
...
+  /// via access(keyPath:) / withMutation(keyPath:) around the private
+  /// backing storage.
+  @ObservationIgnored
   @UserDefaultsBacked("ActiveConnectionId", storage: .group)
-  public var activeConnectionId: UUID? = nil {
-    willSet {
-      objectWillChange.send()
+  private var _activeConnectionId: UUID? = nil
+
+  public var activeConnectionId: UUID? {
```
paulgessinger, line 323: Is this the best way to do this? What are alternatives?

AppShared/Sources/AppShared/Utilities/View+onEvent.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

It's a stand-in for identity comparison. Clarified in `7de7fd2a`.

`onChange(of:)` needs an `Equatable` value, and `Broadcaster` is a non-`Equatable` reference type, so it can't be the observed value directly. `ObjectIdentifier(broadcaster)` is the cheap Equatable/Hashable proxy for "is this literally the same object", so the closure fires exactly when a *different* broadcaster instance gets passed in, and never when an existing one merely emits.

Why it's needed at all: the subscription is created once in `onAppear` and deliberately never torn down on disappear. If the view is later handed a different broadcaster, nothing would notice — events would keep arriving from the old one and the new one would be ignored for the life of the view. This resubscribes instead.

No current call site can hit it: three pass `store.events` and one passes `manager.events`, and `store` is built exactly once (`refreshConnection`'s `else` branch only runs when `store == nil`), so neither identity ever changes. It's here because `onEvent` is public API in a shared package and the failure would be silent.

AppShared/Sources/AppShared/Utilities/View+onEvent.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

It's a one-field mutable slot holding the *latest* handler closure. Expanded the doc comment in `7de7fd2a` — it clearly wasn't carrying its weight at the use site.

The problem it solves: SwiftUI recreates the `ViewModifier` **struct** on every update, and each new struct carries a fresh `action` closure that has captured the current view's properties. But the subscription is created once in `onAppear` and deliberately outlives disappear/reappear — that's the entire reason this modifier exists instead of `.task { for await … }`. So if the subscription captured `action` directly, it would pin the *first* closure for the whole life of the view, and any plain `let`/`var` view property that closure reads would be permanently stale.

The box breaks that: it lives in `@State` (view lifetime, not struct lifetime), `body` overwrites `box.action` on every update, and the subscription calls `box.action?(element)` — one hop, always current. Only the closure is swapped; the subscription and its `Task` are untouched, so no events are dropped.

`.onReceive` gets this for free — its internal `SubscriptionView` is handed the newest closure on every update. Matching that documented shape by hand is what the box is for.

Two things I'd flag as deliberate rather than accidental:
- **Writing to it during `body`** looks like a side effect in a place that shouldn't have one. It's safe specifically because the box is a plain class with no observation: writing to it invalidates nothing, so it cannot loop. Repeated or speculative `body` evaluations just reassign the same closure.
- **`@unchecked Sendable`** — here it's load-bearing, unlike the one I removed from `Subscription`. `action` is written in `body` and read in the sink handler, both main-actor isolated, but that's not expressible to the compiler through a stored `var`.

AppShared/Sources/AppShared/Networking/ConnectionManager.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

It is the right way, and it's the *documented* way — but the comment above it was too vague about why, so I sharpened it in `7de7fd2a`. Leaving the code as-is; alternatives below with what I verified.

**Why the manual pair is forced.** Not a soft "incompatibility": a property wrapper on an `@Observable`-tracked property is a hard compile error. The macro rewrites `x` into computed accessors over generated storage named `_x`, which collides with the wrapper's own `_x`:

```
macro expansion @ObservationTracked:2:6: error: ambiguous reference to member '_x'
```

So the wrapper has to keep the storage, `@ObservationIgnored` has to keep the macro off it, and the public property has to be hand-written. The `access` / `withMutation` pair is exactly what the macro would have generated, and it's what Apple documents for manual observation. Two things about the current code are subtle and correct:

- The key path is the **public** `activeConnectionId`, not `_activeConnectionId`. It has to be keyed on what callers actually read, or observers would never fire. Easy to get backwards.
- `access(keyPath:)` inside its own getter doesn't recurse — it only records the key path.

**Alternatives I looked at:**

1. **Observed stored property + `didSet` write-through.** Keeps the wrapper for encode/decode, drops the manual pair entirely, costs one seeding line in `init`:
   ```swift
   @ObservationIgnored @UserDefaultsBacked("ActiveConnectionId", storage: .group)
   private var _storedActiveConnectionId: UUID? = nil

   public var activeConnectionId: UUID? {
     didSet { _storedActiveConnectionId = activeConnectionId }
   }
   ```
   I verified at runtime that `@Observable` + `didSet` genuinely works — the observation fires (willSet-timed) and `didSet` runs after with the correct `oldValue`. Ergonomically nicer: nothing to forget, no key path to get wrong. The reason I'm **not** recommending it is that it leans on macro/property-observer interaction Apple doesn't document, whereas `access`/`withMutation` is the explicitly blessed escape hatch. Trading a documented pattern for an undocumented one to save two lines isn't a good trade in a store this central.

2. **Teach `UserDefaultsBacked` about Observation** via the `_enclosingInstance` static subscript — the same trick `PublishedUserDefaultsBacked` already uses for `ObservableObject` in that same file. Blocked in practice: the macro generates `access`/`withMutation` as **internal** members of the observable type, so a wrapper in `Common` can't call them on a type in `AppShared` without a protocol re-exposing them. Plus `_enclosingInstance` is underscored, unofficial API — which is exactly why the existing `PublishedUserDefaultsBacked` needs that `@available(*, unavailable) wrappedValue` hack. More machinery, less standard, for one call site.

3. **Move the pointer into the DB.** Your existing comment already rejects this for the right reason and it still holds — the Share Extension gets active-server changes from app-group `UserDefaults` for free.

**One thing worth knowing, not a bug today.** `UserDefaultsBacked.cachedValue` is populated on first read and only invalidated by *local* writes, and nothing in the repo observes `UserDefaults` for external change. So a long-lived process would not see another process's write. That's fine as things stand: I checked, and nothing outside the main app writes `activeConnectionId` — `ShareView.swift:177` only reads it — and the extension is a fresh short-lived process, so its first read is always cold and current. It only becomes a real problem if the extension ever starts writing the pointer. (Side note from the same check: because nothing inside the extension can change that value, `ShareView`'s `.onChange(of: connectionManager.activeConnectionId)` can never fire there. Pre-existing and unrelated to this PR, but you may want to know.)

PR #597, opened 2026-05-31
feat: Element (attribute) cache record models

paulgessinger:
This pull request refactors the `DocumentStore` to centralize and modernize state management by introducing a reactive `ElementStore` projection, simplifying element refresh and CRUD logic, and removing legacy manual cache management. The changes make the store more robust, less error-prone, and easier to maintain by ensuring all element collections are live, read-only projections of the database, and that all updates flow through the repository and observation layer.

**State management modernization:**

* Introduced a new `elementStore` property, which holds a live, reactive projection of all element collections (such as `tags`, `correspondents`, `users`, etc.), replacing the previous manual dictionaries and singletons. All read access to these collections now goes through computed properties that delegate to `elementStore`, ensuring reactivity and consistency. (`DocumentStore.swift`, [AppShared/Sources/AppShared/Document/DocumentStore.swiftL24-R59](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L24-R59))
* Added the `wireElementStore()` method to point the projection at the active repository's database, automatically rewiring on repository changes. (`DocumentStore.swift`, [AppShared/Sources/AppShared/Document/DocumentStore.swiftR93-R111](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R93-R111))
* Removed manual cache clearing and updating logic for element collections; the observation layer now handles all updates and invalidations. (`DocumentStore.swift`, [AppShared/Sources/AppShared/Document/DocumentStore.swiftL129-R168](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L129-R168))

**Element refresh and sync simplification:**

* Replaced all individual `fetchAll*` methods (such as `fetchAllTags`, `fetchAllCorrespondents`, etc.) with a unified `sync(userInitiated:)` method that syncs all elements into the local database, letting the observation repaint the projections. On-demand refreshers now just call `sync`. (`DocumentStore.swift`, [AppShared/Sources/AppShared/Document/DocumentStore.swiftL259-L570](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L259-L570))
* Updated the main `fetchAll()` method to simply trigger a full sync, removing the old concurrent, multi-method fetching logic. (`DocumentStore.swift`, [AppShared/Sources/AppShared/Document/DocumentStore.swiftL259-L570](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L259-L570))

**CRUD operation streamlining:**

* Simplified all create, update, and delete methods for elements (e.g., tags, correspondents, document types, saved views, storage paths) to remove manual cache mutation. Instead, the repository write-throughs to the DB, and the observation layer updates the projections. (`DocumentStore.swift`, [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L259-L570) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L579-L602) [[3]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L647-L679)
* Updated settings mutation logic to work on local copies and write through the repository, since `settings` is now a read-only projection. (`DocumentStore.swift`, [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R459-L622) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L632-R483)

These changes collectively move the codebase to a source-of-truth model with reactive state, reducing the risk of stale or inconsistent UI and making the code easier to reason about and extend.

**Stack** (bottom to top):

<details>

- ~~#597 feat/element-cache-records~~ (merged)  👈

</details>

- #618 feat/multi-server-sync
- #627 feat/background-sync

### DIFF
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -21,24 +21,33 @@ public final class DocumentStore: Sendable {
   // MARK: Observed state
 
   public private(set) var documents: [UInt: Document] = [:]
-  public private(set) var correspondents: [UInt: Correspondent] = [:]
-  public private(set) var documentTypes: [UInt: DocumentType] = [:]
-  public private(set) var tags: [UInt: Tag] = [:]
-  public private(set) var savedViews: [UInt: SavedView] = [:]
-  public private(set) var storagePaths: [UInt: StoragePath] = [:]
-
-  public private(set) var users: [UInt: User] = [:]
-  public private(set) var groups: [UInt: UserGroup] = [:]
-  public private(set) var currentUser: User?
-
-  public private(set) var customFields: [UInt: CustomField] = [:]
-
-  public private(set) var serverConfiguration: ServerConfiguration?
 
   public private(set) var tasks: [PaperlessTask] = []
 
-  public private(set) var permissions: UserPermissions = .empty
-  public private(set) var settings = UISettingsSettings()
+  /// The live element projection (DB → typed `ValueObservation`). The store owns
+  /// it and re-exposes its collections through the computed delegates below, so
+  /// `store.tags` etc. keep working and stay reactive (a view reading
+  /// `store.tags` tracks `ElementStore.tags` through the getter). The reference
+  /// is stable across its lifetime — only its contents change — so it is
+  /// `@ObservationIgnored`; the inner `@Observable` does the tracking.
+  @ObservationIgnored
+  public let elementStore = ElementStore()
+
+  // Element collections and singletons — read-only projections of the DB,
+  // observed via `elementStore`. Writes go through the repository (which
+  // write-throughs to the DB); the observation repaints these.
+  public var correspondents: [UInt: Correspondent] { elementStore.correspondents }
+  public var documentTypes: [UInt: DocumentType] { elementStore.documentTypes }
+  public var tags: [UInt: Tag] { elementStore.tags }
+  public var savedViews: [UInt: SavedView] { elementStore.savedViews }
+  public var storagePaths: [UInt: StoragePath] { elementStore.storagePaths }
+  public var users: [UInt: User] { elementStore.users }
+  public var groups: [UInt: UserGroup] { elementStore.groups }
+  public var customFields: [UInt: CustomField] { elementStore.customFields }
+  public var currentUser: User? { elementStore.currentUser }
+  public var serverConfiguration: ServerConfiguration? { elementStore.serverConfiguration }
+  public var permissions: UserPermissions { elementStore.permissions }
+  public var settings: UISettingsSettings { elementStore.settings }
 
   /// True while a network `sync` is in flight. Distinct from data-presence so a
   /// cold cache shows loading rather than emptiness.
@@ -68,7 +77,6 @@ public final class DocumentStore: Sendable {
   public let events = Broadcaster<Event>()
 
   public let semaphore = AsyncSemaphore(value: 1)
-  public let fetchAllSemaphore = AsyncSemaphore(value: 1)
 
   public private(set) var repository: any Repository
 
@@ -82,10 +90,23 @@ public final class DocumentStore: Sendable {
   public init(repository: some Repository) {
     self.repository = repository
     self.imagePipeline = Self.makeImagePipeline(delegate: repository.delegate)
+    wireElementStore()
   }
 
   deinit {
     taskUpdateTask?.cancel()
+  }
+
+  /// Point the element projection at the active repository's DB. Under the
+  /// source-of-truth model every production/preview repository fronts a DB
+  /// (`CachingBackend`); a repository that doesn't (e.g. `NullRepository` before
+  /// login) detaches the projection.
+  private func wireElementStore() {
+    if let backend = repository as? any CachingBackend {
+      elementStore.repoint(database: backend.database, serverID: backend.serverID)
+    } else {
+      elementStore.reset()
+    }
   }
 
   @Sendable
@@ -135,24 +156,16 @@ public final class DocumentStore: Sendable {
 
   public func clear() {
     documents = [:]
-    correspondents = [:]
-    documentTypes = [:]
-    tags = [:]
-    savedViews = [:]
-    storagePaths = [:]
-    users = [:]
-    groups = [:]
-    currentUser = nil
-    serverConfiguration = nil
     tasks = []
-    permissions = .empty
-    settings = UISettingsSettings()
     lastSyncError = nil
+    // The element projection is owned by `elementStore` and (re)wired by
+    // `wireElementStore()` on repository change — nothing to clear here.
   }
 
   public func set(repository: some Repository, reload: Bool = true) {
     self.repository = repository
     imagePipeline = Self.makeImagePipeline(delegate: repository.delegate)
+    wireElementStore()
     if reload {
       events.emit(.repositoryChanged)
       clear()
@@ -266,174 +279,40 @@ public final class DocumentStore: Sendable {
     await fetchTasks()
   }
 
-  public func fetchAllCorrespondents() async throws {
-    // @TODO: For the `fetchAll` calls: centralize this to that method.
-    //        Use a property on the resource to map to the permissions resource.
-    //        Also: clear the associated resource if there's a permissions error
-    try checkPermission(.view, for: .correspondent)
-    try await fetchAll(
-      elements: repository.correspondents(),
-      collection: \.correspondents)
-  }
+  // On-demand element refreshers kept for their external callers. Under the
+  // source-of-truth model "refresh collection X" means "sync into the DB"; the
+  // live observation then repaints the projection. `syncElements` reconciles
+  // every collection at once, so these all delegate to it.
+  public func fetchAllUsers() async throws { try await sync(userInitiated: true) }
+  public func fetchAllGroups() async throws { try await sync(userInitiated: true) }
+  public func fetchAllCustomFields() async throws { try await sync(userInitiated: true) }
 
-  public func fetchAllDocumentTypes() async throws {
-    try checkPermission(.view, for: .documentType)
-    try await fetchAll(
-      elements: repository.documentTypes(),
-      collection: \.documentTypes)
-  }
-
-  public func fetchAllTags() async throws {
-    try checkPermission(.view, for: .tag)
-    try await fetchAll(
-      elements: repository.tags(),
-      collection: \.tags)
-  }
-
-  public func fetchAllSavedViews() async throws {
-    try checkPermission(.view, for: .savedView)
-    try await fetchAll(
-      elements: repository.savedViews(),
-      collection: \.savedViews)
-  }
-
-  public func fetchAllStoragePaths() async throws {
-    try checkPermission(.view, for: .storagePath)
-    try await fetchAll(
-      elements: repository.storagePaths(),
-      collection: \.storagePaths)
-  }
-
-  public func fetchCurrentUser() async throws {
-    // this should basically always be the case but let's be safe
-    try checkPermission(.view, for: .uiSettings)
-    do {
-      currentUser = try await repository.currentUser()
-    } catch let error where !error.isCancellationError {
-      Logger.shared.error("Unable to get current user: \(error)")
-      throw error
-    }
-  }
-
-  public func fetchAllUsers() async throws {
-    try checkPermission(.view, for: .user)
-    try await fetchAll(
-      elements: repository.users(),
-      collection: \.users)
-  }
-
-  public func fetchAllGroups() async throws {
-    try checkPermission(.view, for: .group)
-    try await fetchAll(
-      elements: repository.groups(),
-      collection: \.groups)
-  }
-
+  /// Refresh `ui_settings` (permissions/settings) from the network. Syncs into
+  /// the DB, then synchronously pulls the singleton into the projection — the
+  /// one path (`DocumentListViewModel.load`) that reads `permissions`
+  /// immediately afterwards can't wait for the observation's runloop hop.
   public func fetchUISettings() async throws {
-    do {
-      // This can fail if we don't have the required permissions to even access UI settings
-      // Older versions of the backend return an ok response here even if the perms aren't valid
-      let uiSettings = try await repository.uiSettings()
-      permissions = uiSettings.permissions
-      settings = uiSettings.settings
-    } catch let error where error.isCancellationError {
-      Logger.shared.debug("Cancelled fetch UI settings")
-    } catch {
-      // If we don't get permissions here, log a warning and assume full permissions.
-      Logger.shared.error("Error getting UI settings: \(error)")
-      Logger.shared.error("Assuming full permissions to proceed")
-      permissions = UserPermissions.full
-      settings = UISettingsSettings()
-      throw error
+    try await sync(userInitiated: true)
+    if let backend = repository as? any CachingBackend {
+      elementStore.refreshUISettings(from: backend.database, serverID: backend.serverID)
     }
   }
 
-  public func fetchAllCustomFields() async throws {
-    try checkPermission(.view, for: .customField)
-    try await fetchAll(
-      elements: repository.customFields(),
-      collection: \.customFields)
-  }
-
-  public func fetchServerConfiguration() async throws {
-    do {
-      serverConfiguration = try await repository.serverConfiguration()
-      let debug = String(describing: serverConfiguration)
-      Logger.shared.info("Fetched server configuration: \(debug, privacy: .public)")
-    } catch let error where error.isCancellationError {
-      Logger.shared.debug("Cancelled fetch server configuration")
-    } catch {
-      Logger.shared.error("Unable to get server configuration: \(error)")
-      throw error
-    }
-  }
-
-  /// Load every element collection through `repository` into the observed dicts.
-  /// Behind a `CachingRepository` this reads the DB (hydrate); behind a plain
-  /// `ApiRepository` it reads the network. `lenient` controls whether a failing
-  /// resource aborts the whole load (`false`, network-fetch semantics) or is
-  /// suppressed (`true`, cache-read semantics — a cold/partial cache must not
-  /// fail the UI).
-  private func loadAll(lenient: Bool) async throws {
-    await fetchAllSemaphore.wait()
-    defer { fetchAllSemaphore.signal() }
-
-    try? await fetchUISettings()
-
-    let permissions = permissions
-    Logger.shared.info(
-      "Permissions returned from backend:\n\(permissions.matrix, privacy: .public)")
-
-    try await withThrowingTaskGroup(of: Void.self) { group in
-      for task in [
-        fetchAllCorrespondents,
-        fetchAllDocumentTypes,
-        fetchAllTags,
-        fetchAllSavedViews,
-        fetchAllStoragePaths,
-        fetchCurrentUser,
-        fetchAllUsers,
-        fetchAllGroups,
-        fetchAllCustomFields,
-        fetchServerConfiguration,
-      ] {
-        group.addTask { try await task() }
-      }
-
-      while !group.isEmpty {
-        do {
-          try await group.next()
-        } catch is PermissionsError {
-          Logger.shared.debug("Load task returned permissions error, suppressing")
-          continue
-        } catch let error where error.isCancellationError {
-          Logger.shared.debug("Load task caught cancellation, suppressing")
-          continue
-        } catch {
-          if lenient {
-            Logger.shared.warning("Load task error (suppressed during hydrate): \(error)")
-            continue
-          }
-          Logger.shared.error("Load task caught error: \(error)")
-          throw error
-        }
-      }
-    }
-  }
-
-  /// Network → DB (behind a `CachingBackend`) or network → dicts (otherwise).
-  /// Automatic syncs fail soft into `lastSyncError`; user-initiated syncs
-  /// rethrow so the caller can surface the failure (toast), as before.
+  /// Network → DB via the caching backend; the live element observation repaints
+  /// the projection. Automatic syncs fail soft into `lastSyncError`;
+  /// user-initiated syncs rethrow so the caller can surface the failure (toast).
   public func sync(userInitiated: Bool = false) async throws {
     Logger.shared.notice("Sync store (userInitiated: \(userInitiated))")
+    guard let backend = repository as? any CachingBackend else {
+      // No DB-backed repository (e.g. NullRepository before login). Nothing to
+      // sync; the projection is empty until a caching repository is set.
+      Logger.shared.info("Sync skipped: repository is not a caching backend")
+      return
+    }
     isRefreshing = true
     defer { isRefreshing = false }
     do {
-      if let backend = repository as? any CachingBackend {
-        try await backend.syncElements()
-      } else {
-        try await loadAll(lenient: false)
-      }
+      try await backend.syncElements()
       lastSyncError = nil
       Logger.shared.info("Sync store complete")
     } catch {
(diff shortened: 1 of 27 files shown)


paulgessinger:
@codex

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -391,20 +404,52 @@ public final class DocumentStore: Sendable {
...
+        Logger.shared.error("Background sync failed (suppressed): \(error)")
+      }
+    }
+  }
+
+  /// The eager entry views call. Triggers a network → DB sync; the cache → UI
+  /// read side is wired up in a later commit layered on top of this core.
+  public func fetchAll() async throws {
+    Logger.shared.notice("Fetch all store request")
+    try await sync()
```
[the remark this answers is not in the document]
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Confirmed, and thanks — this was real, not just theoretical. At the old PR head `loadAll` was unreachable in the app: every connection is wrapped in `CachingRepository`, so `sync()` always took the `CachingBackend` branch and only wrote GRDB. The element dicts, permissions and `currentUser` stayed empty exactly as you describe.

Fixed by **re-cutting the PR boundary** rather than backporting a hydrate. `feat/element-cache-records` now ends at `ada2eb92` (*ElementStore projection — store reads the DB live*), so the cache → UI read path is part of this PR. #597 goes 8 → 11 commits; #598 shrinks correspondingly (its base branch is this one, so its diff re-cut automatically).

The deliberate write-side/read-side split existed because two variants of the offline plan shared the write half — but only the source-of-truth variant proceeded, so the split had stopped paying for itself and left this PR shipping a non-functional app on its own.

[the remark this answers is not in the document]
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Correct, and this one is genuinely latent at this PR — `fetchAll()` → `sync()` defaults to `userInitiated: false`, so the `errorController` catch blocks the picker refreshables already had were dead code.

Tracked as **defer-upstack to #598**, where `4dc776c6` (*surface errors on user-initiated refresh (pull-to-refresh)*) already fixes exactly this: `fetchAll(userInitiated:)` forwards to `sync(userInitiated:)`, the genuinely user-driven paths (`DocumentListViewModel.refresh`, `ManageView.refresh`, the tag and common-picker refreshables) pass `true`, automatic triggers stay `false`, and `DocumentList.refresh()` additionally pushes to the `ErrorController` rather than only logging.

Not backported here — it would rewrite the same lines #598 already touches for no benefit, since the two land together.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -0,0 +1,445 @@
...
+  public func create(shareLink: ProtoShareLink) async throws -> DataModel.ShareLink {
+    try await wrapped.create(shareLink: shareLink)
+  }
+
+  public func delete(shareLink: DataModel.ShareLink) async throws {
+    try await wrapped.delete(shareLink: shareLink)
+  }
+
+  public func update(settings: UISettingsSettings) async throws {
+    try await wrapped.update(settings: settings)
```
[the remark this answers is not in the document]
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Correct — a successful `update(settings:)` left the cached `ui_settings` singleton stale, so relaunching offline restored the old dashboard/sidebar visibility.

Resolved by the PR re-cut described in the P1 thread: `ada2eb92` is now part of this PR, and it writes through exactly as you suggest — merging the new `UISettingsSettings` onto the cached user and permission matrix rather than replacing the record:

```swift
public func update(settings: UISettingsSettings) async throws {
  try await wrapped.update(settings: settings)
  if let current = try database.uiSettings(serverID: serverID) {
    let merged = UISettings(
      user: current.user, settings: settings, permissions: current.permissions)
    try database.setUISettings(merged, serverID: serverID)
  }
}
```

paulgessinger:

## Review round on #597

All three Codex threads were accurate. Two are resolved by a **PR boundary re-cut**, one is deferred to #598. An independent pass over the diff turned up seven further findings; four are fixed here in `94a646f8`, one is held, two were already fixed upstack.

### PR boundary re-cut

Codex's P1 was right that this PR shipped a functionally broken app: `loadAll` was unreachable (every connection is wrapped in `CachingRepository`), so nothing ever populated the store's element dicts, permissions or `currentUser`.

`feat/element-cache-records` was **moved up two commits** to end at `ada2eb92` (*ElementStore projection*). No commits were rewritten — #598's base branch is this one, so both diffs re-cut automatically.

- **#597**: 8 → **11** commits
- **#598**: 11 → **9** commits

### Fixed in `94a646f8`

| | Finding |
|---|---|
| **1** | **`isSkippable` was dead code for all eight paginated collections.** `PageCursor.swift:73` rewrites `RequestError.forbidden` into `ResourceForbidden<Element>` — not a `RequestError` — so `catch let error as RequestError where isSkippable` never matched. A non-superuser lacking `view` on `user`/`group` (a common default) threw out of the sync task group and **permanently killed the entire element sync**, taking tags, correspondents, document types and saved views down with it. `ResourceForbidden` now carries a non-generic `ResourceForbiddenError` marker protocol and `isSkippable` takes `any Error`. This one survived all the way to the top of the stack. |
| **2** | `Database+Elements` bypassed `Database.wrapping(...)` in all eight accessors, leaking raw `GRDB.DatabaseError` out of the Persistence package against the file's own stated contract. |
| **3** | `replaceElements` used `insert`, so an element returned on two consecutive pages — which happens when the server deletes a row mid-pagination — raised a PK violation and rolled back the **whole** reconcile. Now `upsert`. |
| **4** | `lastSyncError = error as? any DisplayableError` was assigned unconditionally, so a non-displayable failure (a GRDB error, a `URLError`) wrote `nil` and **cleared** a degraded state the UI was already showing. |

Findings 2 and 3 compound: before the fix a duplicate row surfaced as an unclassifiable raw GRDB error, which then cleared the degraded-state banner.

### Held for after the stack lands

`CachingRepository` is `@MainActor` and all its SQLite I/O is synchronous, so `syncElements` runs ten serialized write transactions on the main thread; with `busy_timeout = 5000`, a ShareExtension holding the writer lock can stall the main thread for up to five seconds. Architectural rather than a local fix — not worth churning ~60 upstack commits for mid-stack.

### Verification

- New regression test for the `insert`/`upsert` fix. Confirmed it **fails without** the fix with exactly the predicted `UNIQUE constraint failed: tag.server_id, tag.id` — surfacing as `DatabaseError.operationFailed(operation: "replaceElements(tag)")`, which validated the `wrapping(...)` fix in the same run.
- `swift test` green at this PR's head (Persistence 54, Networking 186, Common 54, DataModel 194) and at the stack tip (117 / 203 / 54 / 213).
- `build_sim` (iPhone 17 Pro / iOS 26) green at both this head and the stack tip, zero warnings.
- Inserting the follow-up commit conflicted 59 descendants — all cascades of two hunks (`wtowoqmw`'s `syncUISettings` rewrite, and `votynnqt`'s `Logger.shared` → `Logger.sync` rename landing on the same lines). All resolved in their own commits; the stack is conflict-free.

**Not exercised in the simulator.** Finding 1 in particular needs a real non-superuser account to prove end-to-end, and `CachingRepository` lives in AppShared, which has no test target — so it has no unit test either. Worth a manual check: log in as an account without `view` permission on users/groups and confirm tags and correspondents still populate.

AppShared/Sources/AppShared/Document/ElementStore.swift
```diff
@@ -0,0 +1,203 @@
...
+
+  public private(set) var serverConfiguration: ServerConfiguration?
+
+  /// True once the `ui_settings` singleton has produced a non-nil value for the
+  /// active server — i.e. `permissions`/`settings` reflect real data rather than
+  /// the cold-cache default. Lets the UI distinguish "loading" from "denied".
+  public private(set) var isHydrated = false
+
+  @ObservationIgnored
+  private nonisolated(unsafe) var observationTasks: [Task<Void, Never>] = []
```
paulgessinger: What does `nonisolated(unsafe)` do exactly, and how do we make it safe?

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -0,0 +1,468 @@
...
+    try database.elements(UserGroupRecord.self, serverID: serverID)
+  }
+
+  public func customFields() async throws -> [CustomField] {
+    try database.elements(CustomFieldRecord.self, serverID: serverID)
+  }
+
+  public func currentUser() async throws -> User {
+    guard let user = try database.uiSettings(serverID: serverID)?.user else {
+      throw CachingRepositoryError.cacheMiss
```
paulgessinger, line 194: What's the pattern here? What happens if there's a cache miss on one of these methods?

Persistence/Sources/Persistence/Models/CorrespondentRecord.swift
```diff
@@ -0,0 +1,56 @@
...
+import GRDB
+
+/// GRDB record for a cached `Correspondent` (`correspondent` table).
+public struct CorrespondentRecord: Codable, Sendable, Equatable {
+  public var serverId: UUID
+  public var id: UInt
+  public var name: String
+  public var payload: Payload
+
+  public struct Payload: Codable, Sendable, Equatable {
```
paulgessinger, line 12: Will we potentially be able to change the shape here inside a database migration, or would this require clearing the database and recreating the contents on sync?

Persistence/Sources/Persistence/Models/UISettingsRecord.swift
```diff
@@ -0,0 +1,120 @@
...
+
+  public init(serverId: UUID, domain: UISettings) {
+    self.serverId = serverId
+
+    var permissions = [String: [Bool]]()
+    for resource in UserPermissions.Resource.allCases {
+      permissions[resource.rawValue] = UserPermissions.Operation.allCases.map {
+        domain.permissions.test($0, for: resource)
+      }
+    }
```
paulgessinger, line 74: How does this encode view, change, delete permissions into the string? The ordering in the `[Bool]`?

Persistence/Sources/Persistence/Database+Elements.swift
```diff
@@ -0,0 +1,126 @@
...
+/// mutations).
+extension Database {
+  // MARK: - Multi-row collections
+
+  /// All cached rows of one element kind for a server, ordered by name.
+  public func elements<R: ElementRecord>(
+    _ type: R.Type, serverID: UUID
+  ) throws -> [R.Domain] {
+    try wrapping("elements(\(R.databaseTableName))") {
+      try writer.read { db in
```
paulgessinger, line 36: Reading with the writer? What's that about?

Persistence/Sources/Persistence/Database+Elements.swift
```diff
@@ -0,0 +1,126 @@
...
+        try R
+          .filter(Column("server_id") == serverID)
+          .order(Column("name"))
```
paulgessinger, line 39: The name is indexed so the sorting is reasonably cheap?

Persistence/Sources/Persistence/Database+Elements.swift
```diff
@@ -0,0 +1,126 @@
...
+/// used by sync, which also handles server-side deletes) or a single
+/// write-through (`upsertElement` / `deleteElement`, used by pessimistic
+/// mutations).
+extension Database {
+  // MARK: - Multi-row collections
+
+  /// All cached rows of one element kind for a server, ordered by name.
+  public func elements<R: ElementRecord>(
+    _ type: R.Type, serverID: UUID
+  ) throws -> [R.Domain] {
```
paulgessinger: Could this be a typed throws even? Would that buy us anything? Clarity?

Persistence/Sources/Persistence/Database+Elements.swift
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  /// Replace the entire cached set for a server in one transaction: delete the
+  /// existing rows, insert the new ones. This is how `sync` propagates
+  /// server-side deletions (rows absent from `domains` disappear).
+  ///
+  /// Uses `upsert` rather than `insert` because the caller's paginated fetch can
+  /// legitimately yield the same element twice — deleting a row server-side
+  /// mid-fetch shifts later elements back a page. `insert` raised a primary-key
+  /// violation there and rolled back the whole reconcile.
```
paulgessinger: Don't comment on what it was, just what it is now and why

Persistence/Sources/Persistence/Database+Elements.swift
```diff
@@ -0,0 +1,126 @@
...
+  public func replaceElements<R: ElementRecord>(
+    _ domains: [R.Domain], of type: R.Type, serverID: UUID
+  ) throws {
+    try wrapping("replaceElements(\(R.databaseTableName))") {
+      try writer.write { db in
+        try R.filter(Column("server_id") == serverID).deleteAll(db)
+        for domain in domains {
```
paulgessinger, line 74: There's no good way to batch this I suppose?

Persistence/Sources/Persistence/Database+Observe.swift
```diff
@@ -0,0 +1,78 @@
...
+/// Like ``Database/observeConnections()``, each method wraps GRDB
+/// `ValueObservation` and surfaces it as an `AsyncThrowingStream` of **domain**
+/// values: GRDB, `ValueObservation`, and SQL never cross the `Persistence`
+/// boundary. The first element is the current state (so a consumer's
+/// `for try await` loop is "initial hydrate + live updates" in one), exactly
+/// like the connection observer.
+///
+/// These replace the cache-aside variant's coarse `CacheChange` signal +
+/// `hydrate()`: the observation *carries the data* (the freshly-mapped domain
+/// result), so the store/projection assigns it directly with no re-read.
```
paulgessinger: Don't contrast this to the other explored path, just focus on existing code and reasons

Persistence/Sources/Persistence/Database+Observe.swift
```diff
@@ -0,0 +1,78 @@
...
+///
+/// Scope: in Stage 7 the active server is singular. The observation tracks the
```
paulgessinger: Don't refer to stages in code or comments

AppShared/Sources/AppShared/Document/ElementStore.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

It turns off the compiler's isolation checking for that one property — you assert the safety yourself. It was there for exactly one reason: `ElementStore` is `@MainActor`, but `deinit` is nonisolated and can't read main-actor state.

It was in fact safe, but by an argument the compiler can't see: all mutation is main-actor-only, and `deinit` only runs once the last reference is gone, so nothing can race it.

Made it provably safe instead in `df7a0a1f` — the tasks now live in a small `TaskBox` (a `Sendable` final class over an `OSAllocatedUnfairLock`). `Task.cancel()` is already safe from any thread, so the lock only has to protect the array itself. The annotation is gone.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Two different behaviours. Collections (`tags()`, `users()`, …) return whatever is cached, so a cold cache gives `[]` — indistinguishable from genuinely empty. The three singletons can't do that, because their return type is non-optional, so they throw `CachingRepositoryError.cacheMiss`.

Worth being blunt: **nothing catches `cacheMiss` anywhere**, and it isn't a `DisplayableError` either, so if it did escape it would fail invisibly.

It's unreachable in the app today. The read path is the `ElementStore` observation; `fetchUISettings` goes via `elementStore.refreshUISettings(...)`; and the only app callers of `repository.uiSettings()`/`currentUser()` are in `LoginViewModel`, which builds a bare `ApiRepository` to probe a connection. So these three exist to satisfy the `Repository` protocol. Left as-is, but happy to make that explicit in the code (or give them a displayable error) if you'd rather not rely on that staying true.

Persistence/Sources/Persistence/Models/CorrespondentRecord.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

The payload is JSON in the `data` column, so the schema doesn't constrain its shape. Adding an **optional** field or dropping one is backward-compatible with existing rows.

Adding a **required** field is not, and it fails worse than you might expect: the read is `fetchAll(db).map(\.domain)`, so one undecodable row fails the *entire* collection read, not just that row.

You don't need data-preserving surgery though — these tables hold nothing but derived data. The cheap correct move for a breaking shape change is a forward migration that deletes the affected rows and lets the next sync refill them. Written into the file header in `df7a0a1f` so the next person doesn't reach for JSON rewriting.

Persistence/Sources/Persistence/Models/UISettingsRecord.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Yes — the ordering in the `[Bool]` *is* the encoding. The dictionary is keyed by `Resource.rawValue` (`"documenttype"`, `"storagepath"`, …), and the array is positional by `Operation`, which is `Int`-raw and declared `view, add, change, delete` → 0,1,2,3. Encode uses `Operation.allCases.map`, decode uses `flags[operation.rawValue]`; they agree only because `allCases` order matches raw-value order.

The sharp edge is that a future `Operation` may only be **appended**: decode bounds-checks and treats a missing trailing flag as `false`. Inserting one in the middle shifts every raw value and makes already-cached rows decode as the *wrong permissions*, silently. Given this is the permission matrix, that seemed worth writing down — noted in the type's doc comment in `df7a0a1f`, along with the fact that such a change needs a migration clearing `ui_settings`.

Persistence/Sources/Persistence/Database+Elements.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Naming artifact rather than a real serialization. `writer` is `any DatabaseWriter`, and `DatabaseWriter` refines `DatabaseReader` — so `writer.read` on the production `DatabasePool` takes a *reader* connection and runs concurrently with writes under WAL. It only serializes on the in-memory `DatabaseQueue` used in tests, which is single-connection by construction. The property name is what makes it look wrong. Noted in the file header in `df7a0a1f`.

Persistence/Sources/Persistence/Database+Elements.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

It wasn't — good catch, the premise was false. V3 created only the implicit `(server_id, id)` primary-key index, so this was a scan plus a sort, re-run on every observation emission (and `replaceElements` rewrites the whole table each sync, which re-fires the observation).

Fixed in `df7a0a1f`. Note a plain `name` index wouldn't have helped — the read filters by server *and* orders by name, so it needs the composite **`(server_id, name)`**. V3 hasn't shipped (main carries only V1 and V2), so the index went into the original migration rather than a V4.

Measured rather than assumed, via `EXPLAIN QUERY PLAN`. Before:

```
SEARCH tag USING INDEX sqlite_autoindex_tag_1 (server_id=?)
USE TEMP B-TREE FOR ORDER BY
```

After, the temp b-tree is gone and the read walks `index_tag_on_server_id_name`. Two tests added: one asserting every multi-row table carries the index, one asserting the ordered read produces no temp b-tree. Both verified to fail without the index.

Persistence/Sources/Persistence/Database+Elements.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Yes, and it buys more than clarity — done in `df7a0a1f`.

Since every accessor now funnels through `wrapping(...)`, `DatabaseError` really is the only thing they can throw, so `throws(DatabaseError)` is expressible on all 9 plus `wrapping` itself. The payoff is that it makes that invariant **compiler-enforced** rather than a convention: an accessor that reaches GRDB directly no longer compiles.

Verified by deliberately breaking one:

```
error: thrown expression type 'any Error' cannot be converted to error type 'DatabaseError'
```

That's precisely the bug fixed by hand earlier in this round (raw GRDB errors escaping the package), so it's now unrepresentable. It erases at the `Repository` boundary, which declares untyped `throws` — so there's no call-site win, only an internal one.

Persistence/Sources/Persistence/Database+Elements.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Fixed in `df7a0a1f` — it now states the rule and the reason, with no archaeology:

```swift
/// `upsert` rather than `insert`: a paginated fetch can legitimately deliver
/// the same element twice, because deleting a row server-side mid-fetch
/// shifts later elements back a page. The last write for an id wins, and the
/// reconcile survives the duplicate.
```

Applied the same edit to the other two comments I wrote in the previous round that had the same problem (`isSkippable` and the `lastSyncError` guard) rather than only the line you flagged.

Persistence/Sources/Persistence/Database+Elements.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Not really worth it. GRDB caches the prepared statement per record type, so the loop is already statement-reuse inside a single transaction, and the transaction overhead is amortised across the batch. A hand-built multi-row `VALUES (…),(…)` would save marginal per-row dispatch in exchange for hand-written SQL and a parameter-count cap to manage. Left as the loop.

Persistence/Sources/Persistence/Database+Observe.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Removed in `df7a0a1f`. The paragraph now describes what the observation does rather than what it isn't:

```swift
/// Each emission carries the freshly-mapped domain result, so a consumer
/// assigns it directly and never re-reads the database to find out what
/// changed.
```

The matching mention in `ElementStore`'s header comment got the same treatment.

Persistence/Sources/Persistence/Database+Observe.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Removed in `df7a0a1f` — all 4 occurrences in the files this PR adds, not just this one (`Database+Observe`, `V3_CreateElementCache` ×2, and the `MARK` in `CachingRepository`).

Two notes. The `CachingRepository` one collided with an upstack commit that rewrites that section with its own stage reference; I resolved it by taking the upstack text minus the reference, so it's fixed there too.

And there are **8 more** in files this PR doesn't touch — `DataModel` models (`Stage 3 principle` ×4), `ApiSavedView`, `Broadcaster`, and `Database+Connections` ×2. Those are already on `main`, so I left them out of scope here. Say the word and I'll do a separate sweep — including the upstack ones, of which `CachingRepository` alone has 5 more by the top of the stack.

PR #598, opened 2026-05-31
feat: Element (attribute) cache is source-of-truth

paulgessinger:
This pull request introduces several improvements to the `DocumentStore` and `CachingRepository` to enhance data synchronization, permission enforcement, and preview/testability. The main changes ensure that network syncs are coalesced, permission checks are enforced for all mutations, and preview stores can be more easily constructed for testing. Additionally, the UI settings now gate which collections are fetched during sync, and localization strings for "Tap for details!" have been simplified.

### Data Synchronization and Performance

* Added a `syncTask` to `DocumentStore` to ensure that concurrent sync requests join a single in-flight network sync, preventing redundant network calls and improving efficiency (`DocumentStore.swift`) [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R88-R94) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L282-R324) [[3]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R341-R373).
* The UI settings permission matrix is now fetched first in `CachingRepository.syncElements()`, gating which collections are fetched based on user permissions. If fetching UI settings fails, the sync falls back to cached permissions or fetches all collections as before (`CachingRepository.swift`) [[1]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR72-R130) [[2]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feL124-R162).

### Permission Enforcement

* All create, update, and delete operations in `DocumentStore` now require explicit permission checks for the relevant resource before proceeding, ensuring that only authorized actions are performed (`DocumentStore.swift`) [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R383-R387) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R398-R410) [[3]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R427-R462) [[4]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R471-R493) [[5]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R546-R554) [[6]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R563-R579).

### Preview/Testability

* Added a `preview(_:)` static method and `previewRepository(as:)` instance method to `DocumentStore`, allowing for easy construction of a preview store with an in-memory seeded database for UI previews and testing (`DocumentStore+Preview.swift`).
* Exposed the wrapped repository in `CachingRepository` for better testability (`CachingRepository.swift`).

### Logging and Safety

* Improved logging and added a safety check in `DocumentStore.set(repository:)` to warn if a non-caching repository is installed while the app is live, preventing UI from silently emptying due to detachment from the database (`DocumentStore.swift`).

### Localization

* Simplified the "Tap for details!" localization string to just "Details" (and equivalents in other languages), making the UI more concise (`App.xcstrings`).

---

## Manual test checklist

Derived from a pass over all 17 commits in this PR. Nothing here has been driven in a
simulator or on device; `AppShared` has no test target, so none of it has automated
coverage either.

### Element mutations under restricted permissions — *new in this PR*

Document and note mutations were already permission-checked on `main`; this PR extends
that to the five element types, so any mistake shows up as a previously-working action
that now refuses.

- [x] On a **non-superuser account with mixed permissions**, create / rename / delete a **tag** — permitted operations still work, denied ones fail with a permission message rather than a generic error
- [x] Same for **correspondent**, **document type**, **storage path**
- [x] Same for **saved view** (create, update, delete — including the visibility handling on create)
- [x] The inline-creation paths specifically: "+" in the tag picker, and the correspondent / document-type / storage-path edit sheets on a document

### Permission errors name the operation — *added after the first test pass*

A refusal used to read "No access: Storage path" whichever operation was attempted, and a
server-side 403 didn't even say that much ("The request was denied due to missing
permissions"). Both now surface as a `PermissionsError` naming the operation. The toast
message is one line with tail truncation, so the operation is the *title* and the resource
moved into the details.

- [x] A refused **delete** of a storage path toasts **"Delete not permitted"**, and **Details** names the resource ("… not allowed to delete Storage path"); check a refused **create** and **edit** too
- [x] The title **fits on one line** without truncation — including one longer non-English locale if you have one handy
- [x] The same wording appears for a **server-side 403** — an element the account can list but not modify (object-level permissions), where the client-side matrix says the operation is allowed. The backend's own message is appended under **Details**
- [x] **Document** and **note** mutations (delete document, add / delete note) still surface their failures, now with the operation named
- [x] Spot-check one non-English locale: the eight new strings are English-only until Crowdin runs, so they should fall back rather than show a key
- [x] *Known, not fixed here:* after a **failed** delete in a manage list the row stays hidden until you refresh — [#665](https://github.com/paulgessinger/swift-paperless/issues/665)

### Permissions refresh on document detail — *added after the first test pass*

The detail view's pull-to-refresh reloaded the document and its metadata but never
`ui_settings`, so a permission change on the server didn't reach that screen until the next
launch or list refresh.

- [x] Open a document, change the account's permissions server-side, **pull to refresh** on the document — the edit / delete affordances follow the new permissions without restarting the app
- [x] The refresh still loads document and metadata as before, and a failed `ui_settings` fetch logs without toasting over the document

### Cold start and offline permission hydration — *highest risk, fixed after the first test pass*

Confirmed during testing: an offline cold start read as **"no permissions"** everywhere,
because `permissions` was `.empty` until `ui_settings` landed and `.empty` tests false for
everything. `ElementStore.isHydrated` existed to tell "loading" from "denied" and had no
readers (ledger [#664](https://github.com/paulgessinger/swift-paperless/issues/664)).

Now `DocumentStore.permissions` hands out **`.full`** until `permissionsKnown`, so gates
fall through to the request and the server decides. Two places still consult the flag:
`userCan*(document:)` (which also depend on `currentUser`, nil while unhydrated) and
`PermissionsView`, which *displays* the matrix and must not report access the server never
granted.

- [x] **First-ever launch against a server while offline** (use the debug **Clear Cached Data** action from #600 to reproduce, which keeps servers and wipes the caches): lists and documents stay usable, nothing reads as "Insufficient permissions"
- [x] **Relaunch offline with a warm cache**: elements still populate from the DB and mutations are still permitted per the cached matrix
- [x] **Settings → server → Permissions during a cold start**: every cell shows **?**, with the "not loaded" note — *not* an all-green matrix. Copy summary is prefixed with `(not loaded from server; assuming full access)`
- [x] Once `ui_settings` arrives the matrix flips to the real values, and a genuinely denied screen shows its lock again
- [x] *Behaviour change:* create / edit / delete affordances are now **enabled** during a cold start instead of greyed out. Offline, using one should fail with a **connectivity** error, not a permission error
- [x] The document detail no longer claims "you don't have global change permission" while permissions are still loading

### Permission-gated sync

`syncElements()` now fetches `ui_settings` first and skips collections the matrix says the
user can't view.

- [x] Log in as an account **without `view` on users/groups** and confirm tags, correspondents, document types, storage paths and saved views still populate (this is ledger #656 item 4's check — the gating it depends on is added here)
- [x] Confirm the permissions editor's owner / view / edit pickers still list users and groups on an account that *is* allowed to see them

### Share Extension — *now DB-cached and needs-auth wrapped*

It previously ran on a bare `ApiRepository`; it's now `CachingRepository(NeedsAuthRepository(ApiRepository))`, so element data comes from its own cache and 401s are intercepted.

- [x] Share a file from another app: tag / correspondent / document type / storage path pickers populate
- [x] Create a **new tag** from inside the extension and confirm the document saves with it
- [x] With an **expired or invalid token**, confirm the extension surfaces re-auth instead of failing opaquely — this interception didn't exist here before
- [x] *Expected, not a bug:* an element created in the extension does **not** reach an already-running app until the next **successful** foreground sync (ledger #656 item 13)

### Error presentation

- [x] Open **Settings**, trigger an error, tap **Details** on the toast — the alert appears **above** the sheet and the sheet stays open. Previously presenting it tore the sheet down
- [x] The toast action now reads **"Details"** (was "Tap for details!") and the generic title **"Unexpected error"** (was "An unexpected error occurred") — spot-check one non-English locale

### Sync behaviour

- [x] **Pull-to-refresh while offline** on the document list now raises an error toast; it was silently swallowed before
- [x] Opening the **users/groups pickers** or the **custom fields** screen while offline no longer toasts — these were mislabeled as user-initiated and now fail soft
- [x] Launch fires **one** element sync rather than several — `Joining in-flight element sync` should appear in the log for the callers that piggy-back
- [x] **Connection switch mid-sync:** with two servers configured, start a sync against a slow or unreachable one and switch while it's still running. The new server's elements should populate on the switch, not only after a later foreground or pull-to-refresh
- [x] Interrupt a **pull-to-refresh** with a connection switch — no error toast should appear

### Extra headers on a live connection

- [x] Edit extra headers on the active connection and confirm tags/correspondents/etc. survive (ledger #656 item 6, never verified; this PR adds the fault log that would catch a regression)

### Xcode previews

- [x] `FilterBar` preview shows the Inbox tag, correspondent, document type and storage path
- [x] Spot-check one or two other `TransientRepository`-backed previews (e.g. `TagFilterView`, `TrashView`)

### Baseline smoke

- [x] Fresh install → login → every element collection populates
- [x] Force-quit and relaunch **offline** → elements still shown from the cache

**Stack** (bottom to top):

<details>

- ~~#598 feat/element-cache-source-of-truth~~ (merged)  👈

</details>

- #618 feat/multi-server-sync
- #627 feat/background-sync

### DIFF
--- a/AppShared/Sources/AppShared/Document/DocumentStore+Preview.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore+Preview.swift
@@ -1,0 +1,53 @@
+//
+//  DocumentStore+Preview.swift
+//  AppShared
+//
+//  Preview/test convenience for building a `DocumentStore` under the
+//  source-of-truth model, where every repository must front a DB for the
+//  element projection to see anything.
+//
+
+import Foundation
+import Networking
+import Persistence
+
+extension DocumentStore {
+  /// A store backed by an in-memory seeded DB, wrapping `wrapped` in a
+  /// `CachingRepository` so the live `ElementStore` projection and the
+  /// write-through mutations behave exactly as in production.
+  ///
+  /// The wrapped repository's element data is copied into the DB by a one-shot
+  /// `fetchAll()` (previews are live, so it lands a beat after first render);
+  /// `PreviewRepository`'s built-in fixtures appear this way. Previews using a
+  /// `TransientRepository` seed through `store.repository` (writes flow to the
+  /// DB) and recover the underlying repository for its non-`Repository` helpers
+  /// via ``previewRepository(as:)``.
+  @MainActor
+  public static func preview(_ wrapped: some Repository = PreviewRepository()) -> DocumentStore {
+    let serverID = UUID()
+    let database: Database
+    do {
+      database = try Database.seeded(serverID: serverID)
+    } catch {
+      // The in-memory seed (DatabaseQueue + migrations) is infallible in
+      // practice; a preview crash here is loud and immediately actionable.
+      preconditionFailure("Preview database seed failed: \(error)")
+    }
+    let caching = CachingRepository(wrapping: wrapped, database: database, serverID: serverID)
+    let store = DocumentStore(repository: caching)
+    Task { try? await store.fetchAll() }
+    return store
+  }
+
+  /// Recover the underlying repository from a store built by ``preview(_:)`` —
+  /// for previews that need a concrete repository's preview-only helpers (e.g.
+  /// `TransientRepository.addUser`/`login`/`allDocuments`). The protocol
+  /// surface should be used through `store.repository` so writes reach the DB.
+  @MainActor
+  public func previewRepository<R: Repository>(as type: R.Type) -> R {
+    guard let caching = repository as? CachingRepository<R> else {
+      preconditionFailure("previewRepository(as:) called on a store not built by preview(_:)")
+    }
+    return caching.wrapped
+  }
+}
--- a/AppShared/Sources/AppShared/Document/DocumentStore+Preview.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore+Preview.swift
@@ -17,7 +17,7 @@ extension DocumentStore {
   /// write-through mutations behave exactly as in production.
   ///
   /// The wrapped repository's element data is copied into the DB by a one-shot
-  /// `fetchAll()` (previews are live, so it lands a beat after first render);
+  /// `sync()` (previews are live, so it lands a beat after first render);
   /// `PreviewRepository`'s built-in fixtures appear this way. Previews using a
   /// `TransientRepository` seed through `store.repository` (writes flow to the
   /// DB) and recover the underlying repository for its non-`Repository` helpers
@@ -35,7 +35,7 @@ extension DocumentStore {
     }
     let caching = CachingRepository(wrapping: wrapped, database: database, serverID: serverID)
     let store = DocumentStore(repository: caching)
-    Task { try? await store.fetchAll() }
+    Task { try? await store.sync() }
     return store
   }
 
--- a/AppShared/Sources/AppShared/Document/DocumentStore+Preview.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore+Preview.swift
@@ -7,6 +7,7 @@
 //  element projection to see anything.
 //
 
+import DataModel
 import Foundation
 import Networking
 import Persistence
@@ -22,12 +23,25 @@ extension DocumentStore {
   /// `TransientRepository` seed through `store.repository` (writes flow to the
   /// DB) and recover the underlying repository for its non-`Repository` helpers
   /// via ``previewRepository(as:)``.
+  ///
+  /// The `ui_settings` singleton is seeded with a full permission matrix rather
+  /// than left to that sync: every store mutation is permission-checked, and the
+  /// sync cannot be what supplies the matrix. It is fire-and-forget, and for a
+  /// `TransientRepository` it fails outright — `uiSettings()` rethrows
+  /// `noUserLoggedIn` until the preview logs a user in from its own `.task`,
+  /// which happens after this sync has already run and found no cached row to
+  /// fall back to. A preview that seeds through `store.create(…)` would lose
+  /// every element to `PermissionsError`. The matrix is applied synchronously so
+  /// it holds from the first render, not a runloop hop later; a later sync
+  /// overwrites it with the repository's own settings.
   @MainActor
   public static func preview(_ wrapped: some Repository = PreviewRepository()) -> DocumentStore {
     let serverID = UUID()
     let database: Database
     do {
-      database = try Database.seeded(serverID: serverID)
+      database = try Database.seeded(
+        serverID: serverID,
+        uiSettings: UISettings(user: previewUser, permissions: .full))
     } catch {
       // The in-memory seed (DatabaseQueue + migrations) is infallible in
       // practice; a preview crash here is loud and immediately actionable.
@@ -35,9 +49,14 @@ extension DocumentStore {
     }
     let caching = CachingRepository(wrapping: wrapped, database: database, serverID: serverID)
     let store = DocumentStore(repository: caching)
+    store.elementStore.refreshUISettings(from: database, serverID: serverID)
     Task { try? await store.sync() }
     return store
   }
+
+  /// Stands in for `ui_settings.user` until a sync supplies the real one. Matches
+  /// the user `Database.seeded` puts on the preview connection record.
+  private static let previewUser = User(id: 1, isSuperUser: true, username: "preview")
 
   /// Recover the underlying repository from a store built by ``preview(_:)`` —
   /// for previews that need a concrete repository's preview-only helpers (e.g.
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -298,16 +298,6 @@ public final class DocumentStore: Sendable {
     await fetchTasks()
   }
 
-  // On-demand element refreshers kept for their external callers. Under the
-  // source-of-truth model "refresh collection X" means "sync into the DB"; the
-  // live observation then repaints the projection. `syncElements` reconciles
-  // every collection at once, so these all delegate to it. They are automatic
-  // on-appear refreshers, not user gestures, so `userInitiated: false` — a
-  // failure fails soft (cached data stays on screen) rather than toasting.
-  public func fetchAllUsers() async throws { try await sync() }
-  public func fetchAllGroups() async throws { try await sync() }
-  public func fetchAllCustomFields() async throws { try await sync() }
-
   /// Refresh `ui_settings` (permissions/settings) from the network, then
   /// synchronously pull the singleton into the projection — the one path
   /// (`DocumentListViewModel.load`) that reads `permissions` immediately
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -87,7 +87,7 @@ public final class DocumentStore: Sendable {
 
   // The in-flight element sync, if any. Concurrent `sync` callers join this one
   // task instead of each firing their own `syncElements` (the launch flurry of
-  // fetchAll / fetchUISettings / scenePhase triggers shares a single network
+  // sync / fetchUISettings / scenePhase triggers shares a single network
   // pass). Only ever touched on the main actor.
   @ObservationIgnored
   private var syncTask: Task<Void, Error>?
@@ -317,7 +317,8 @@ public final class DocumentStore: Sendable {
   /// `userInitiated` policy to the shared outcome — automatic syncs fail soft
   /// into `lastSyncError`, user-initiated syncs rethrow so the caller can
   /// surface the failure (toast). So a user-initiated call joining a background
-  /// sync still sees the error.
+  /// sync still sees the error. This is what entry views call eagerly on
+  /// appear, and what pull-to-refresh calls with `userInitiated: true`.
   public func sync(userInitiated: Bool = false) async throws {
     Logger.shared.notice("Sync store (userInitiated: \(userInitiated))")
     do {
@@ -361,16 +362,6 @@ public final class DocumentStore: Sendable {
       isRefreshing = false
     }
     try await task.value
-  }
-
-  /// The eager entry views call. Triggers a network → DB sync; the element
-  /// projection repaints from the live observation. `userInitiated` forwards to
-  /// `sync`: pass `true` for explicit refreshes (pull-to-refresh) so failures
-  /// rethrow and the caller can surface them; automatic triggers (launch,
-  /// foreground) leave it `false` to fail soft into `lastSyncError`.
-  public func fetchAll(userInitiated: Bool = false) async throws {
-    Logger.shared.notice("Fetch all store request (userInitiated: \(userInitiated))")
-    try await sync(userInitiated: userInitiated)
   }
 
   public func document(id: UInt) async throws -> Document? {
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -233,50 +233,54 @@ public final class DocumentStore: Sendable {
 
   public func updateDocument(_ document: Document) async throws -> Document {
     Logger.shared.info("Updating document with ID \(document.id, privacy: .public)")
-    try checkPermission(.change, for: .document)
-    events.emit(.changed(document: document))
+    return try await performing(.change, on: .document) {
+      events.emit(.changed(document: document))
 
-    var document = document
+      var document = document
 
-    if settings.documentEditing.removeInboxTags {
-      Logger.shared.debug("Removing inbox tags from document as per setting")
-      let inboxTags = tags.values.filter(\.isInboxTag)
-      for tag in inboxTags {
-        document.tags.removeAll(where: { $0 == tag.id })
+      if settings.documentEditing.removeInboxTags {
+        Logger.shared.debug("Removing inbox tags from document as per setting")
+        let inboxTags = tags.values.filter(\.isInboxTag)
+        for tag in inboxTags {
+          document.tags.removeAll(where: { $0 == tag.id })
+        }
       }
-    }
 
-    let updated = try await repository.update(document: document)
-    documents[updated.id] = updated
-    events.emit(.changeReceived(document: updated))
-    return updated
+      let updated = try await repository.update(document: document)
+      documents[updated.id] = updated
+      events.emit(.changeReceived(document: updated))
+      return updated
+    }
   }
 
   public func deleteDocument(_ document: Document) async throws {
     Logger.shared.info("Deleting document with ID \(document.id, privacy: .public)")
-    try checkPermission(.delete, for: .document)
-    try await repository.delete(document: document)
-    documents.removeValue(forKey: document.id)
-    events.emit(.deleted(document: document))
+    try await performing(.delete, on: .document) {
+      try await repository.delete(document: document)
+      documents.removeValue(forKey: document.id)
+      events.emit(.deleted(document: document))
+    }
   }
 
   public func deleteNote(from document: Document, id: UInt) async throws {
     Logger.shared.info("Deleting note with ID \(id, privacy: .public)")
-    try checkPermission(.delete, for: .note)
-    events.emit(.changed(document: document))
-    _ = try await repository.deleteNote(id: id, documentId: document.id)
+    try await performing(.delete, on: .note) {
+      events.emit(.changed(document: document))
+      _ = try await repository.deleteNote(id: id, documentId: document.id)
 
-    events.emit(.changeReceived(document: document))
+      events.emit(.changeReceived(document: document))
+    }
   }
 
   public func addNote(to document: Document, note: ProtoDocument.Note) async throws {
     Logger.shared.info("Adding note to document \(document.id, privacy: .public)")
-    try checkPermission(.add, for: .note)
-    events.emit(.changed(document: document))
+    try await performing(.add, on: .note) {
+      events.emit(.changed(document: document))
 
-    _ = try await repository.createNote(documentId: document.id, note: note)
+      _ = try await repository.createNote(documentId: document.id, note: note)
 
-    events.emit(.changeReceived(document: document))
+      events.emit(.changeReceived(document: document))
+    }
   }
 
   public func notes(for document: Document) async throws -> [Document.Note] {
@@ -394,13 +398,14 @@ public final class DocumentStore: Sendable {
     method: (E) async throws -> R
   ) async throws -> R
   where E: Sendable & PermissionsModel, R: Identifiable & Sendable {
-    try checkPermission(.add, for: resource)
-    // `settings` is kept live by the element observation, so its permission
-    // defaults are already current — apply them directly. The repository
-    // write-throughs the created element to the DB; the observation repaints it
-    // into the projection.
-    let updated = settings.permissions.appliedAsDefaults(to: element)
-    return try await method(updated)
+    try await performing(.add, on: resource) {
+      // `settings` is kept live by the element observation, so its permission
+      // defaults are already current — apply them directly. The repository
+      // write-throughs the created element to the DB; the observation repaints
+      // it into the projection.
+      let updated = settings.permissions.appliedAsDefaults(to: element)
+      return try await method(updated)
+    }
   }
 
   private func update<E>(
@@ -408,8 +413,9 @@ public final class DocumentStore: Sendable {
     resource: UserPermissions.Resource,
     method: (E) async throws -> E
   ) async throws where E: Identifiable & Sendable {
-    try checkPermission(.change, for: resource)
-    _ = try await method(element)
+    try await performing(.change, on: resource) {
+      _ = try await method(element)
+    }
   }
 
   private func delete<E>(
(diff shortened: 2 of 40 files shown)


AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -329,11 +338,39 @@ public final class DocumentStore: Sendable {
...
-  public func fetchAll() async throws {
-    Logger.shared.notice("Fetch all store request")
-    try await sync()
+  /// projection repaints from the live observation. `userInitiated` forwards to
+  /// `sync`: pass `true` for explicit refreshes (pull-to-refresh) so failures
+  /// rethrow and the caller can surface them; automatic triggers (launch,
+  /// foreground) leave it `false` to fail soft into `lastSyncError`.
+  public func fetchAll(userInitiated: Bool = false) async throws {
+    Logger.shared.notice("Fetch all store request (userInitiated: \(userInitiated))")
+    try await sync(userInitiated: userInitiated)
```
paulgessinger: This is now a pure alias? Can we drop it then?

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -56,7 +56,7 @@ enum CachingRepositoryError: Error {
 
 @MainActor
 public final class CachingRepository<Wrapped: Repository>: Repository, CachingBackend {
-  private let wrapped: Wrapped
+  let wrapped: Wrapped
```
paulgessinger, line 63:
```suggestion
  private(set) let wrapped: Wrapped
```
?

ShareExtension/ShareView.swift
```diff
@@ -14,34 +14,15 @@ import os
...
-    return ConnectionManager(database: database)
-  }()
-
+  // The extension process's own Database (app-group SQLite, WAL). The same DB
+  // backs the `ConnectionManager` and — wrapped in a `CachingRepository` in
+  // `refreshConnection` — the element cache, so the store's `ElementStore`
+  // projection observes the extension's own writes. Cross-process live
+  // notification isn't delivered (the extension syncs at launch), but the
+  // extension's in-process writes drive its own observation normally.
+  private let database: Database
```
paulgessinger, line 23: Ok, so we don't have delivery of changes across the process boundary. Makes sense. What's our plan though in case the user creates an attribute in the share extension? When will it show up in the main app? Next launch?

ShareExtension/ShareView.swift
```diff
@@ -54,6 +35,28 @@ struct ShareView: View {
...
+    do {
+      return try Database()
+    } catch {
+      Logger.shared.fault(
+        "Share Extension database bootstrap failed (\(error)); falling back to in-memory")
+      if let inMemory = try? Database.inMemory() {
+        return inMemory
+      }
+      preconditionFailure(
+        "In-memory database fallback also failed; cannot construct ConnectionManager")
```
paulgessinger: This hard crashes? Still log so we can possibly diagnose a report?

paulgessinger:
🤖 _Posted by Claude on Paul's behalf._ This review was produced by an agent reading the
code at this PR's head (`feat/element-cache-source-of-truth`), not at the stack tip. Nothing
below was driven in the simulator — findings are from code reading plus tracing call sites;
where a claim is inferred rather than confirmed by execution, it's marked. CI is green on all
8 checks.

---

### 1. `syncTask` isn't invalidated on a repository swap — the new server's first sync is silently skipped

**Verified by code reading.** `AppShared/Sources/AppShared/Document/DocumentStore.swift:344`

`runSyncElements()` joins any non-nil `syncTask` *before* it looks at `repository`, and
`set(repository:)` (`DocumentStore.swift:172`) never clears it. The connection-switch path in
`swift_paperlessApp.swift:171` is exactly `store.set(repository:)` → `try? await store.fetchAll()`.

Scenario: a launch/foreground sync against server A is still running (nine collections plus
server config, with serialized main-thread SQLite writes — see #656 item 2 — so the window is
seconds, not milliseconds). The user switches connection. `set(repository:)` installs B's
`CachingRepository`; the following `fetchAll()` logs *"Joining in-flight element sync"*, awaits
**A's** task, sets `lastSyncError = nil` and returns success. The `.repositoryChanged` event →
`viewModel.reload()` → `load()` → `fetchUISettings()` → `sync()` joins the same stale task. B's
tags/correspondents/users are never fetched, nothing surfaces an error, and the projection stays
empty until the next scenePhase sync.

Suggested fix, in `set(repository:)`:

```swift
syncTask?.cancel()
syncTask = nil
```

If that goes in, the owner's teardown at `DocumentStore.swift:359` needs guarding too — otherwise
the *old* owner's `defer { syncTask = nil }` clears a task a newer caller installed, and the caller
after that starts a duplicate concurrent sync:

```swift
defer {
  if syncTask === task { syncTask = nil }
  isRefreshing = false
}
```

(That second half is latent today — no new owner can install a task while `syncTask` is non-nil —
but it becomes reachable the moment the swap clears it.)

### 2. A user-initiated refresh can join a stale sync and return without a fresh network pass

**Inferred (design-level), not reproduced.** `DocumentStore.swift:345`

Coalescing is unconditional, so a pull-to-refresh or *Sync now* issued a few seconds into an
existing background sync gets that sync's result rather than a new pass. The doc comment covers
error propagation ("a user-initiated call joining a background sync still sees the error") but not
staleness — the user pulls and nothing refetches. Worth either skipping the join when
`userInitiated`, or setting a dirty flag so a follow-up sync runs once the in-flight one lands.

### 3. An empty permission matrix silences the entire element sync

**Verified by code reading; reachability inferred.** `AppShared/Sources/AppShared/Networking/CachingRepository.swift:79`

`gate?.test(.view, for: resource) ?? true` — the `nil` fallback only covers `uiSettings()`
*throwing*. A response that omits `permissions` decodes to `.empty`
(`ApiUISettings.domain`: `permissions?.domain ?? .empty`), and `UserPermissions.test` returns
`false` for an absent rule, so `canView` is false for all eight collections and nothing syncs —
with `lastSyncError = nil`. Same for a cached `.empty` matrix.

Blast radius is limited because every UI read site already gates on that same matrix, but this
turns "affordances hidden" into "cache never fills, silently". Cheap hardening: treat an all-false
matrix like `nil` and fall back to the per-resource 403 skip `syncCollection` already has.

### 4. Two call sites lost their error surfacing

**Verified by code reading.** `AppShared/Sources/AppShared/Views/CustomFields/CustomFieldEditView.swift:145`,
`AppShared/Sources/AppShared/Views/PermissionsEditView.swift` (`initialize()`)

`fetchAllCustomFields()` / `fetchAllUsers()` were `sync(userInitiated: true)` and rethrew; the
replacements call bare `store.sync()`, which fails soft. `CustomFieldEditView`'s
`catch { errorController.push(error:) }` is therefore unreachable — a failed custom-field refresh
is now silent. `PermissionsEditView.fetch` only logs, so that one is cosmetic. Both look like they
want `sync(userInitiated: true)`.

### 5. Minor

- `checkPermission` on element mutations makes a cold matrix fail closed locally: on a fresh
  install, before the first `uiSettings` sync, creating a tag throws `PermissionsError`
  ("no permission") rather than a network error. Probably fine — worth confirming the message
  reads sensibly in that state.
- `DocumentStore.preview`'s fire-and-forget `Task { try? await store.fetchAll() }` races previews
  that seed `TransientRepository` in their own `.task`. It's one-shot, so a late seed shows nothing.
  Acknowledged in the doc comment; previews only.
- The `App.xcstrings` change hand-edits `translated` values in seven non-English locales
  (`Tap for details!` → `Details`, etc.). Those are Crowdin-managed and the edits will be clobbered
  on the next pull.

---

### Re-read requested by #656 item 8 — resolved

That item asks for a second pair of eyes on `wtowoqmw`'s conflict resolution (the `syncUISettings`
rewrite, which "took the upstack version, dropping the now-targetless `isSkippable` edit").

**The resolution is correct.** #597's widened `isSkippable` — the `error is any ResourceForbiddenError`
clause that was the F1 fix — is intact at `CachingRepository.swift:174` and still used by both
`syncCollection` and `syncServerConfiguration`. `syncUISettings` dropped its own use of it only
because it now catches *everything* and returns the cached matrix, which is strictly more tolerant.
Nothing was lost in the merge.

---

### Review round of 2026-08-09 — dispositions

🤖 _Appended by Claude on Paul's behalf._ Paul's four inline comments, and what each one
became. Three follow-up commits sit on top of this PR's head; the whole stack was rebased
over them and re-pushed.

| Comment | Disposition | Result |
|---|---|---|
| [`fetchAll` is a pure alias — drop it?](https://github.com/paulgessinger/swift-paperless/pull/598#discussion_r3745054369) | apply-now | [`d0abc135`](https://github.com/paulgessinger/swift-paperless/commit/d0abc135de9a) — removed; all 22 call sites now call `store.sync(...)`. [Reply](https://github.com/paulgessinger/swift-paperless/pull/598#discussion_r3745188715) |
| [`private(set) let wrapped`](https://github.com/paulgessinger/swift-paperless/pull/598#discussion_r3745062981) | reject | Doesn't compile (`'private(set)' modifier cannot be applied to read-only properties`). A `let` is already read-only; only the access level was widened, for `previewRepository(as:)`. Documented in [`75124e0b`](https://github.com/paulgessinger/swift-paperless/commit/75124e0b1757). [Reply](https://github.com/paulgessinger/swift-paperless/pull/598#discussion_r3745188932) |
| [When does a Share-Extension attribute reach the app?](https://github.com/paulgessinger/swift-paperless/pull/598#discussion_r3745075689) | hold (post-stack) | Next foreground, via the server — `scenePhase == .active` → `store.sync()` rewrites the rows and fires the app's own observation. The offline hole (write sits unread in the shared DB) is tracked as **#656 item 13**. [Reply](https://github.com/paulgessinger/swift-paperless/pull/598#discussion_r3745189756) |
| [`preconditionFailure` without a log](https://github.com/paulgessinger/swift-paperless/pull/598#discussion_r3745076935) | apply-now | [`ad53ab55`](https://github.com/paulgessinger/swift-paperless/commit/ad53ab5539bf) — the in-memory fallback is a `do`/`catch` now; its error gets a `fault` log and is interpolated into the precondition message. [Reply](https://github.com/paulgessinger/swift-paperless/pull/598#discussion_r3745189985) |

**Effect on the findings above.** They were written against `730c9a51`; `fetchAll` no longer
exists, so finding 1's trigger path is now `set(repository:)` → `store.sync()` (the defect is
unchanged — `syncTask` still isn't cleared on a repository swap), and every `file:line`
reference has shifted by the follow-up commits. Findings 1–5 are otherwise untouched: none
of them were addressed in this round.

Verification of the round: both this PR's head and the stack tip build clean for the iPhone 17
Pro simulator, no revision between them still references `fetchAll`, and no commit in the stack
is left conflicted. Still nothing driven in the simulator.

**Re-review note.** Rebasing the rename conflicted three upstack commits, resolved in place, so
#600, #618 and #627 present slightly different diffs than before.

AppShared/Sources/AppShared/Document/DocumentStore.swift

Dropped in `d0abc135`. It was a pure forward (log line + `sync(userInitiated:)`), so all 22 call sites now call `store.sync(...)` directly and `sync`'s doc comment absorbed the "entry views call this eagerly" framing.

Rebasing it through the stack conflicted 3 commits (`DocumentListViewModel`, `swift_paperlessApp` ×2) — all resolved in the conflicted commits themselves. Verified: no revision from here to the stack tip still references `fetchAll`, and both this PR's head and the tip build clean for the simulator.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Not applicable, unfortunately — `private(set)` on a `let` is a compile error:

```
error: 'private(set)' modifier cannot be applied to read-only properties
```

(verified with `swiftc -typecheck`). A `let` is already read-only to everyone, including this module; the only thing the diff changed was the *access level*, from `private` to internal, so `DocumentStore.previewRepository(as:)` can recover the wrapped repository.

Added a doc comment saying exactly that in `75124e0b`, so the next reader doesn't have to ask.

ShareExtension/ShareView.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Next foreground, not next launch — and via the server rather than via the shared DB.

The app re-syncs on `scenePhase == .active` (`swift_paperlessApp.swift` → `store.sync()`). That sync re-fetches the elements from the server and rewrites the same rows in the DB, which *does* fire the app's own in-process observation, so the tag the extension created shows up there. No relaunch needed.

The hole is the offline case: the extension's write is sitting right there in the shared app-group DB, but nothing re-reads it, so if the foreground sync fails the app stays blind to it. A DB re-read on foreground (`ElementStore.repoint(database:serverID:)`) would close that regardless of network outcome.

Not doing it here — it touches `ElementStore` and the scenePhase path, both of which several upstack PRs rework. Tracked as item 13 in #656.

ShareExtension/ShareView.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Yes, it hard-crashes — deliberately, since a `ConnectionManager` needs a `Database` and there is no third fallback. But you're right that it was crashing without the evidence: `try?` threw the error away and `preconditionFailure` only carries its message into the report.

Fixed in `ad53ab55` — the fallback is a `do`/`catch` now, the error gets a `fault` log of its own, and it's interpolated into the precondition message too so it lands in the crash report even if the log is gone.

paulgessinger:
@codex

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -329,11 +339,29 @@ public final class DocumentStore: Sendable {
...
-  public func fetchAll() async throws {
-    Logger.shared.notice("Fetch all store request")
-    try await sync()
+  /// Run (or join) the single in-flight `syncElements`. Returns when it
+  /// completes; throws its error to every joined caller (who each decide what to
+  /// do with it). A no-op without a caching backend.
+  private func runSyncElements() async throws {
+    if let syncTask {
+      Logger.shared.debug("Joining in-flight element sync")
+      return try await syncTask.value
```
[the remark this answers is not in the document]
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Confirmed and fixed in `ca92cfef`.

`set(repository:)` now cancels and clears `syncTask` before installing the new repository, so the follow-up `sync()` can't join the outgoing backend's task. Both call sites do exactly the sequence you describe — `swift_paperlessApp.swift:171-173` and `ShareExtension/ShareView.swift:92-94`. The extension's window is the wider of the two: `refreshConnection()` fires from `.onAppear` *and* `.onChange(of: connections)`, so a hydrate-driven switch can land mid-sync.

Two consequences of the fix that weren't obvious up front:

- **`runSyncElements`'s teardown had to become identity-guarded** (`if syncTask == task`). Cancelling on replacement makes two syncs able to own `syncTask` in sequence, and without the guard the cancelled sync's `defer` would clear the *replacement's* handle — breaking coalescing — and drop `isRefreshing` while the replacement was still running. That second-order bug didn't exist before, since only a task's own creator could ever clear the handle.
- **`sync` no longer rethrows cancellation to `userInitiated` callers.** Otherwise a switch interrupting a pull-to-refresh would surface a bare `CancellationError`: `DocumentListViewModel.surface` pushes whatever it's handed. I checked all 28 `sync()` call sites at the stack tip — the only other error-surfacing one is that same path, and `BackgroundSyncCoordinator` already uses `try? await store.sync()` — so nothing depended on the throw.

Verified: #598's head and the stack tip both build clean for the simulator, and the fix still lands correctly inside `install(repository:reload:)` after #618's refactor (both `activate` callers still do `activate` → `sync()`).

Not verified: never driven in the simulator. `AppShared` has no test target, so there's no automated coverage either — staging the race needs two servers and a slow or unreachable connection.

AppShared/Sources/AppShared/Document/DocumentStore+Preview.swift
```diff
@@ -0,0 +1,53 @@
...
+    do {
+      database = try Database.seeded(serverID: serverID)
+    } catch {
+      // The in-memory seed (DatabaseQueue + migrations) is infallible in
+      // practice; a preview crash here is loud and immediately actionable.
+      preconditionFailure("Preview database seed failed: \(error)")
+    }
+    let caching = CachingRepository(wrapping: wrapped, database: database, serverID: serverID)
+    let store = DocumentStore(repository: caching)
+    Task { try? await store.sync() }
```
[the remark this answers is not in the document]
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Confirmed and fixed in `f4f47a58` — though it's worse than "asynchronous fire-and-forget leaves the store with `.empty` permissions". For `TransientRepository` previews it isn't a race at all; it fails deterministically:

1. `preview(_:)` fires its sync while the repository still has no logged-in user.
2. `uiSettings()` rethrows `currentUser()`'s `noUserLoggedIn` (`TransientRepository.swift:380,556`).
3. `syncUISettings` swallows that and falls back to the cached row — of which a fresh in-memory DB has none — returning `nil` (`CachingRepository.swift:157-166`). **No `ui_settings` row is ever written.**
4. `permissions` stays `.empty`, and `UserPermissions.test` returns `false` for any resource with no rule.
5. FilterBar's `.task` calls `repository.login(userId: 1)` and *then* `store.create(tag:)` (`FilterBar.swift:983-992`) → `PermissionsError` → bare `catch {}`. All four seed elements silently lost.

Step 5 is why the "affected previews must await a sync before mutating" half of the suggestion wouldn't have worked: the preview's own `try await store.sync()` sits *after* its creates, and the login happens after `preview(_:)` has already returned, so no amount of awaiting the initial sync helps.

So the fix takes the other half — seed synchronously. `Database.seeded` already accepts a `uiSettings:` argument, so `preview(_:)` seeds a full matrix through it and then pulls it into the projection with `elementStore.refreshUISettings`, which makes previews permission-hydrated before the first render rather than a runloop hop later. A later sync overwrites it with the repository's real settings.

One correction to the scope: `PreviewRepository`-backed previews were only ever exposed to the narrow race you described — its `uiSettings()` returns `UserPermissions.full` with no login required (`PreviewRepository.swift:476-482`), and their mutations are user-driven.

Verified: builds clean at #598's head and at the stack tip. Not verified: the previews themselves weren't opened in Xcode to watch the seed elements appear.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -121,12 +150,20 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
     }
   }
 
-  private func syncUISettings() async throws {
+  /// Fetch the UI settings singleton and return its permission matrix to gate
+  /// the rest of the sync. Never throws: on any failure it falls back to the
+  /// last cached matrix, or `nil` if none exists (caller then fetches every
+  /// collection and relies on per-resource 403/401-skip, as before). A
+  /// uiSettings failure therefore degrades gating without aborting the sync.
+  private func syncUISettings() async -> UserPermissions? {
```
paulgessinger, line 158: 🤖 _Posted by Claude on Paul's behalf._

**`syncUISettings` no longer throwing removes the only signal a `ui_settings` failure had.**

This came out of re-reviewing the #597 rebase conflict resolutions (ledger #656 item 8). The resolution here took the upstack version, which is genuinely *safer* than the one it displaced — it catches everything rather than only 403/401, so the `isSkippable` edit it dropped really was targetless. The problem is the signature change that rode along with it.

On `main` (post-#597) this was `async throws` and caught only skippable errors, so a 500, a decode error or a transport failure propagated out of the task group and aborted `syncElements()`. Here it is `async -> UserPermissions?` and catches everything.

That throw was load-bearing. `DocumentStore.sync(userInitiated:)` rethrows when the user pulled to refresh, so on `main` a failing `ui_settings` surfaced as a toast. Now the failure reaches nothing:

- **the log** — `Logger.sync.info`, invisible at the level exported logs are read at;
- **`lastSyncError`** — only fills from a *thrown* `DisplayableError`, and nothing throws now;
- **Offline & Sync** — `syncErrors` is `QuerySyncError`, i.e. per-saved-view fill failures only, and only rendered when the server is in *Entire library* mode.

So a server persistently 500ing on `ui_settings` leaves the app running on a stale-or-absent permission matrix with no indication at all — and that matrix feeds the element mutation permission checks this same PR adds, so create/edit affordances can appear or vanish with no explanation. Not data loss, but a silent functional degradation where there used to be a toast.

**Suggested fix** — keep the non-throwing signature (the gating design needs it) but split the catch:

- skippable (403/401) → `.info`, as now;
- everything else → `Logger.sync.error`, plus an error the UI can actually reach.

**Related, same file:** `syncServerConfiguration` still catches only skippable, so a 500 *there* aborts the whole sync and skips `setLastSyncAt`. Two singletons in one sync with opposite failure policies — worth making that a deliberate choice rather than leaving it as a rebase artefact.

The wider pattern — six failure paths in this stack all logging at `.info`, with no user-visible surface behind them — is tracked as ledger #656 item 15. Fixing this one does not address the other five.

<sub>**Verified** by reading the code at this PR's head (`f4f47a58`) and at `main` (`d47af410`): the signature and catch clause on both sides, `DocumentStore.sync(userInitiated:)`'s `userInitiated` rethrow, `syncErrors`'s type and its `mode == .entireLibrary` render condition, and the log levels at each site. **Not verified:** nothing was run in a simulator or on device. The "there used to be a toast" claim is read off the code path rather than observed, and the consequence for the permission affordances is reasoning from the permission checks this PR adds, not something reproduced.</sub>

PR #600, opened 2026-06-02
feat: Document cache source-of-truth

paulgessinger:
This pull request introduces a major refactor to the document caching and synchronization logic, implementing a robust offline-first architecture for document lists and details. The changes remove the in-memory document cache from `DocumentStore`, shifting all state to the database and introducing a new cache observation and fill API. Additionally, it adds mechanisms for reconciling remote deletions and metadata changes, ensuring local data stays consistent with the server, and improves pessimistic write-through and cache fallback behavior for document operations.

**Document caching and observation API:**

* Removed the in-memory `documents` dictionary from `DocumentStore`, and replaced all access with direct observation of the database-backed cache. Added new methods such as `documentQueryKey`, `fillDocumentQuery`, `observeDocumentPrefix`, `observeQueryStatus`, and `observeDocument` to provide a GRDB-free, offline-first interface for views. [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L23-L24) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R670-R765)

* Implemented a full document list fill (`fillQuery`) in `CachingRepository`, which eagerly fetches the first page and then background-fills the rest, enabling instant offline access and smooth scrolling. [[1]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR64-R83) [[2]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR182-R219)

**Remote reconciliation and cache consistency:**

* Added throttled remote-delete reconciliation (`reconcileDocuments` and `reconcileDocumentDeletions`), which periodically fetches the server's authoritative document IDs and removes any locally cached documents that no longer exist on the server. Also added a delta-based changed-metadata refresh (`reconcileDocumentChanges`) to keep cached documents up-to-date without re-fetching entire lists. [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R116-R123) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R367-R370) [[3]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR64-R83) [[4]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feL366-R647)

* Introduced `DocumentDeltaWatermark`, a per-server metadata watermark stored in UserDefaults, to efficiently track and limit delta refreshes. Also provided a method to clear all watermarks when wiping the local cache. [[1]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR32-R52) [[2]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feL366-R647)

**Write-through and fallback improvements:**

* Updated document update, delete, and fetch methods in `CachingRepository` to write through confirmed changes to the cache, mark queries as stale, and serve cached data on network failures, ensuring consistent offline behavior and immediate UI updates.

**Other enhancements:**

* Added cache wipe logic to `DocumentStore`, ensuring all local data (database, blobs, image cache, and watermarks) can be cleared for maintenance or debugging.
* Extended repository protocols and implementations to support efficient ID fetching and fallback strategies. [[1]](diffhunk://#diff-89d221ed691ac94c47dbd34fd50c6fd0b087608b1fb700dbc2d9b5e60856aa6cR105-R108) [[2]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feL366-R647)

These changes collectively enable robust, efficient, and user-friendly offline document browsing and editing, while keeping the local cache tightly synchronized with the server.

---

**Document caching and observation:**
- Removed in-memory document cache from `DocumentStore` and replaced with database-backed, observable cache API for document lists and details. Added new methods for query key computation, cache fill, observation, and cache wipe. [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L23-L24) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R670-R765)
- Implemented `fillQuery` in `CachingRepository` for eager, background-filled document list caching, enabling instant offline access. [[1]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR64-R83) [[2]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR182-R219)

**Remote reconciliation and cache consistency:**
- Added throttled remote-delete reconciliation (`reconcileDocuments`/`reconcileDocumentDeletions`) and delta-based metadata refresh (`reconcileDocumentChanges`) to keep the local cache consistent with server state. [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R116-R123) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R367-R370) [[3]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR64-R83) [[4]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feL366-R647)
- Introduced `DocumentDeltaWatermark` for efficient per-server metadata sync, and a method to clear all watermarks during cache wipe. [[1]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR32-R52) [[2]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feL366-R647)

**Write-through and fallback improvements:**
- Updated document update, delete, and fetch operations to write through to the cache, mark affected queries as stale, and serve cached data on network failures for robust offline-first behavior.

**Other enhancements:**
- Added comprehensive cache wipe logic to `DocumentStore` for database, blobs, image cache, and per-server watermarks.
- Extended repository protocols for efficient document ID fetching and fallback. [[1]](diffhunk://#diff-89d221ed691ac94c47dbd34fd50c6fd0b087608b1fb700dbc2d9b5e60856aa6cR105-R108) [[2]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feL366-R647)

## Review round (2026-08-16)

Six follow-up commits at the end of this PR's range, from an agent-authored review pass
(details and the full finding list in [this comment](https://github.com/paulgessinger/swift-paperless/pull/600#issuecomment-5306727572); deferrals tracked in #656).

| Commit | What it fixes |
|---|---|
| `keep a document from being placed twice in one cached query` | `query_order` had no uniqueness on `remote_id`, and the in-memory dedup this PR removed was never replaced — so a document the server repeats across adjacent pages was stored twice and rendered twice. Unique key with `ON CONFLICT IGNORE`, plus `.insert` in place of `.upsert`. |
| `let the list view model deallocate while its observations run` | `guard let self` outside an infinite `for await` kept the view model and its two `ValueObservation` streams alive forever; each iPad rotation stacked another pair. |
| `don't let a position collision abort a query fill` | Follow-on from the first commit: `.insert` made a `position` collision throw where `upsert` had silently overwritten, and the background fill swallows the throw. Explicit `.replace` on the primary key. |
| `page the remote-delete id sweep over a unique ordering` | The sweep inherited `FilterState.empty`'s sort — archive serial number, NULL for most documents — so paging it was unstable, and anything dropped across a page boundary reads as a remote deletion. |
| `send the negotiated search API on the document-id sweep` | `documentIDs` always spelled its rules the legacy way while its neighbour passed the negotiated value. |
| `report a failed blob purge instead of claiming success` | `purge()` swallowed the removal error and then recreated the directory, so a failed wipe reported clean — and the debug action exists to leave an *empty* store for cold-launch testing. |

Two findings are deliberately **not** fixed here:

- **An interrupted list fill truncates the cache and reports it as complete** — the highest-impact
  finding of the round, deferred to #618 where the per-query error surface exists (#656 item 18).
  This PR ships with it live.
- **The changed-metadata delta stranded any backlog larger than its cap** — fixed on #617 instead,
  where that function has its final shape.

Verified: builds clean at the stack tip, all four packages pass, no conflicted commit in the stack.
Nothing was driven in the simulator.

## Review round 2 (2026-08-16, Paul's inline comments)

Twenty inline comments; six more follow-up commits here, plus one on #617 and one on #627.

| Commit | What it does |
|---|---|
| `reach the document cache through Database.wrapping` | Every document-cache accessor was untyped `throws` calling GRDB directly, against the invariant `Database+Elements` states — so raw `GRDB.DatabaseError`s escaped the package. All of `Database+Documents` / `+DocumentDetail` / `+Maintenance` now declare `throws(DatabaseError)`. No caller changes needed. |
| `create server_sync_state in the migration that first uses it` | `V4` created a table nothing in this build reads or writes; both its columns belong to #617. Moved to #617's `V6` (`270489a9`), where the first reader lives. |
| `express the order-stale marking in GRDB's query interface` | Raw `UPDATE … WHERE query_key IN (SELECT …)` replaced by a subquery + `updateAll`. |
| `scope DocumentStore.clear and name the metadata version lookup` | `clear()` had no callers outside the type → `private`, with its now-much-smaller job stated. The dense version-id optional chain became `fileMetadataVersionID(documentId:)`. |
| `say what a cached query's total actually drives` | Eight places claimed `totalCount` was a "scrollbar extent". It drives the count pill; the list only renders the loaded prefix. Also trimmed the id-sweep comment. |
| `page the documentIDs default over a unique ordering too` | The `Repository` default implementation carried the same non-unique-paging bug the `ApiRepository` override was fixed for last round. |

Filed from this round: **#669** — cached query keys are never garbage-collected when they stop being reachable, which blocks both changing `FilterState.empty`'s default sort and retiring `QueryKey`'s `.legacy` encoding. Ledger item 10 (stale `Stage N` comments) is closed by a sweep at the top of #627.

Verified: all four packages pass (629 tests), app builds clean at this PR's head *and* at the stack tip, no conflicted commit in the stack.

## Review round 3 (2026-08-17, lint pass)

Two follow-up commits at the end of this PR's range.

| Commit | What it does |
|---|---|
| `apply swift-format to the queryStatus signature` | CI lint was red. Last round's `reach the document cache through Database.wrapping` added `throws(DatabaseError)` to `Database.queryStatus`, pushing the opening brace onto its own line; swift-format joins it back. Confirmed against the version CI actually pins (602.0.0, via prek) — the local toolchain is 603.0.0 and agrees. |
| `drop the documentIDs default, make the paging fallback explicit` | Answers the one open inline comment on `CachingRepository.documentIDs`. The `Repository` default that paged the full Tier-1 list is gone; `documentIDs` is now a bare requirement, and the paging fallback survives as a helper (`pagedDocumentIDs(filter:)`) that `NullRepository` / `PreviewRepository` / `TransientRepository` opt into by name. |

The second commit changes the answer to that thread. The passthrough on `CachingRepository`
was previously deletable-but-load-bearing: removing it still compiled and silently dropped
callers onto the paging default instead of `ApiRepository`'s `fields=id` projection. With the
default gone, deleting it is a compile error — the footgun the comment was pointing at no
longer exists. Keeping it on the protocol (rather than moving it to an extension) remains
required: `CachingRepository<Wrapped>` calls it through a protocol-constrained generic, so a
non-requirement would dispatch statically and miss the override.

Still open from that thread: whether to point `reconcileDocumentDeletions` at `self.documentIDs`
instead of `wrapped.documentIDs`, so the passthrough is genuinely exercised. Behaviour-neutral,
not done here.

**Stack impact.** #617–#627 were rebased onto this PR's new head. Two real conflicts in
`Repository.swift`, both where #617 adds `documents(filter:fullPerms:)` immediately next to
`documentIDs` and then removes it again a few commits later: resolved by keeping both in
`full_perms list projection (R4b)`, and by applying the removal in `always request full_perms on
the document list`. The other 37 conflicted commits were propagation and cleared with them.

Verified: `Networking` builds clean at all five PR heads, 215 Networking tests pass at the stack
tip, the app scheme builds clean on the simulator at this PR's head *and* at the stack tip, and
prek is green here. Nothing was driven in the simulator.

Not fixed: `current_changelog.txt` loses its trailing newline from #617's `Offline & Sync settings
screen` commit onward, so `end-of-file-fixer` fails on #617/#618/#620/#627 (not on this PR).
Pre-existing and unrelated to this round.

## Manual test checklist

Nothing in this PR has been driven in the simulator — all verification so far is builds, unit
tests, and lint. This is the list that would actually exercise it. Roughly ordered so each
section leaves the app in a useful state for the next.

**Setup:** a server with enough documents to page (≥ 100), at least one saved view, and a way to
change data server-side (web UI in a browser next to the simulator).

### 1. Cold start and first fill

- [x] Settings → Advanced → **Clear Cached Data**, confirm. Toast reads "Cached data cleared", server connections survive.
- [x] Force-quit, relaunch. Document list appears without a visible empty-state flash before the first rows land.
- [x] The count pill shows the server's total, not just the number of rows loaded so far.
- [x] Scroll to the bottom. Rows keep arriving; no gaps, no spinner that never resolves.
- [x] Scroll the whole list looking for the **same document appearing twice** — this was a real bug (`keep a document from being placed twice in one cached query`) and page boundaries are where it showed.

### 2. Offline browsing

- [x] With the list populated, enable Airplane Mode. Back out to the list and return — rows still render from cache.
- [ ] Pull to refresh while offline. Expect **no** error message: `ErrorController.suppressBannerCoveredErrors` drops connectivity-class errors while the device is offline by design. The `offlineToast` it defers to only fires on the online→offline *transition*, so once that has gone there is no surface at all — filed as **#671**, pre-existing and not introduced here.
- [x] Open a document you viewed while online: title, tags, correspondent, and **notes** all render offline.
- [x] Open the metadata sheet for that document — cached file metadata is there.
- [x] Open a document you have *not* opened before while offline. It should fail gracefully, not hang on a loading overlay.
- [x] Leave Airplane Mode. The list recovers on its own or on one pull-to-refresh.

### 3. Detail view load behaviour

- [x] Open a document, back out, reopen. Reopening is instant and shows **no loading spinner** (loads on appear are deliberately silent now).
- [ ] Pull to refresh *inside* the detail view. This one *does* show progress. **Known broken at this PR's head** — every refresh cancels itself and toasts "cancelled", including with no new document version; fixed in #620 by `859716b1b` (moves the load off the gesture's task), which should cover any trigger, not just the version bump its commit message describes. Explicit re-test owed there: [#620 comment](https://github.com/paulgessinger/swift-paperless/pull/620#issuecomment-5319956615).
- [x] Confirm the thumbnail stays visible while the full preview loads — it should not blank out and come back.
- [x] Open a document with several notes. They should be there immediately on second open (prefetched).

### 4. Mutations write through

- [x] Edit a document's title. The list row updates immediately, without a manual refresh.
- [x] Change tags from the tags sheet and save. Row reflects it.
- [ ] Now break it on purpose: Airplane Mode on, change tags, save. The save fails **and the tags visibly revert** rather than sticking in the changed-but-unsaved state.
- [x] Delete a document in-app. It disappears from the list and does not come back on refresh.

### 5. Reconciling with server-side changes

- [x] Delete a document **in the web UI**, then pull to refresh in the app. It disappears locally. (This is the unfiltered id sweep; it's throttled, so give it a refresh or two.)
- [x] Delete several documents server-side at once and refresh. All of them go — none survive, and *nothing else* disappears with them. (Unstable paging here used to read as false deletions.)
- [x] Edit a document's title **in the web UI**, refresh in the app. The change shows up without re-fetching the whole list.
- [x] Switch to a saved view and back to the default list. Both render correct, distinct contents.
- [x] Use search/filters, then clear them. Results are correct and the list returns to the full set.

### 6. Scanner and odd paths

- [x] Scan or enter an ASN that exists. It resolves to the right document (this PR added the index behind that lookup).
- [x] Scan an ASN that doesn't exist — clean "not found", no hang.
- [x] On iPad: open the list and **rotate 10–15 times**, then keep using the app. Watch for slowdown or memory growth. A leak here kept a view model and two DB observations alive per rotation; it's fixed, but rotation is how you'd see a regression.

### 7. Cache wipe

- [x] **Clear Cached Data** again with documents cached and a preview downloaded. Confirm the toast, then check that documents re-download and the app is not left in a broken state.
- [ ] Verify server connections are still configured afterwards.

### Known-live, please don't file these

- **No feedback when an action fails offline** — pull-to-refresh says nothing, and a tag save offline shows a spinner, returns the save button, keeps your edit, and says nothing (the change is *not* saved). Deliberate suppression whose surface only fires on the offline *transition*. Filed as **#671** (ledger #656 item 28); pre-existing, not introduced here.

- **Pull-to-refresh scroll hitch** — the scroll stops tracking your finger for a beat at the trigger threshold, then snaps to catch up. Main-thread SQLite: #660 (also noted on #669 as a possible contributing factor). Reproduces offline as well, slightly weaker. Accepted for now.

- **An interrupted list fill truncates the cache and reports it as complete** — deferred to #618 (#656 item 18). If you kill the app mid-fill and the list is short until the next refresh, that's this.
- **Cached query keys are never garbage-collected** (#669). Not user-visible; storage only.

**Stack** (bottom to top):

<details>

- ~~#600 feat/document-cache-source-of-truth~~ (merged)  👈

</details>

- #618 feat/multi-server-sync
- #627 feat/background-sync

### DIFF
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -113,6 +113,14 @@ public final class DocumentStore: Sendable {
   @ObservationIgnored
   private var syncTask: Task<Void, Error>?
 
+  // Last successful (or attempted) remote-delete reconcile. The sweep fetches
+  // the server's whole id set, so it is throttled — `sync()` kicks it
+  // fire-and-forget on launch/foreground/refresh, but it only runs at most once
+  // per `reconcileThrottle` unless user-initiated.
+  @ObservationIgnored
+  private var lastDocumentReconcile: Date?
+  private let reconcileThrottle: TimeInterval = 300
+
   // MARK: Methods
 
   public init(repository: some Repository) {
@@ -356,6 +364,10 @@ public final class DocumentStore: Sendable {
       try await runSyncElements()
       lastSyncError = nil
       Logger.shared.info("Sync store complete")
+      // Reconcile remote deletes alongside the element sync (throttled,
+      // non-blocking). Pull-to-refresh (userInitiated) bypasses the throttle.
+      let userInitiated = userInitiated
+      Task { [weak self] in await self?.reconcileDocuments(userInitiated: userInitiated) }
     } catch {
       // A cancellation is never the user's problem to see: either the caller's
       // own task went away, or `set(repository:)` retired this sync on a
@@ -695,6 +707,24 @@ extension DocumentStore {
   ) -> AsyncThrowingStream<QueryStatus, Error> {
     guard let backend = repository as? any CachingBackend else { return Self.emptyStream() }
     return backend.database.observeQueryStatus(queryKey: queryKey, serverID: backend.serverID)
+  }
+
+  /// Throttled remote-delete reconcile: drop cached documents that no longer
+  /// exist on the server (so they disappear from every offline list). Soft-fail
+  /// (background); kicked from `sync()`. `userInitiated` bypasses the throttle.
+  public func reconcileDocuments(userInitiated: Bool = false) async {
+    guard let backend = repository as? any CachingBackend else { return }
+    if !userInitiated, let last = lastDocumentReconcile,
+      Date().timeIntervalSince(last) < reconcileThrottle
+    {
+      return
+    }
+    lastDocumentReconcile = Date()
+    do {
+      try await backend.reconcileDocumentDeletions()
+    } catch {
+      Logger.shared.info("Document reconcile failed (suppressed): \(error)")
+    }
   }
 
   /// Live single document by id, for detail/preview surfaces that must repaint on
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -663,6 +663,14 @@ public final class DocumentStore: Sendable {
 /// `NullRepository` before login) the observations are empty, finished streams
 /// and `fillDocumentQuery` throws.
 extension DocumentStore {
+  /// The stable key for a list query, computed without touching the network, so
+  /// the list can subscribe to whatever is already cached (offline-first) before
+  /// — or independently of — the network fill. `nil` without a caching backend.
+  public func documentQueryKey(filter: FilterState) -> QueryKey? {
+    guard let backend = repository as? any CachingBackend else { return nil }
+    return QueryKey(serverID: backend.serverID, filter: filter)
+  }
+
   /// Eager full-fill of a list: await page 1 + count, then background-page the
   /// rest. The returned handle carries the `QueryKey` the list then observes.
   public func fillDocumentQuery(filter: FilterState) async throws -> QueryFillHandle {
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -20,8 +20,6 @@ import os
 public final class DocumentStore: Sendable {
   // MARK: Observed state
 
-  public private(set) var documents: [UInt: Document] = [:]
-
   public private(set) var tasks: [PaperlessTask] = []
 
   /// The live element projection (DB → typed `ValueObservation`). The store owns
@@ -180,16 +178,13 @@ public final class DocumentStore: Sendable {
     taskUpdateTask = Task(operation: taskPoller)
   }
 
-  public func clearDocuments() {
-    documents = [:]
-  }
-
   public func clear() {
-    documents = [:]
     tasks = []
     lastSyncError = nil
-    // The element projection is owned by `elementStore` and (re)wired by
-    // `wireElementStore()` on repository change — nothing to clear here.
+    // Documents are not held in memory under source-of-truth — the list observes
+    // the DB directly (see the document-cache surface below). The element
+    // projection is owned by `elementStore` and (re)wired by `wireElementStore()`
+    // on repository change — nothing to clear here.
   }
 
   public func set(repository: some Repository, reload: Bool = true) {
@@ -269,8 +264,9 @@ public final class DocumentStore: Sendable {
         }
       }
 
+      // The repository write-throughs the confirmed value to the DB; the list/detail
+      // observations repaint it in place (no in-memory dict to update).
       let updated = try await repository.update(document: document)
-      documents[updated.id] = updated
       events.emit(.changeReceived(document: updated))
       return updated
     }
@@ -279,8 +275,9 @@ public final class DocumentStore: Sendable {
   public func deleteDocument(_ document: Document) async throws {
     Logger.shared.info("Deleting document with ID \(document.id, privacy: .public)")
     try await performing(.delete, on: .document) {
+      // The repository write-throughs the delete to the DB; the FK cascade removes
+      // it from every cached query_order and the observations repaint.
       try await repository.delete(document: document)
-      documents.removeValue(forKey: document.id)
       events.emit(.deleted(document: document))
     }
   }
@@ -655,6 +652,52 @@ public final class DocumentStore: Sendable {
       Logger.api.debug("No permissions for \(operation.description) on \(resource.rawValue)")
       throw PermissionsError(resource: resource, operation: operation)
     }
+  }
+}
+
+// MARK: - Document cache surface (Stage 8)
+
+/// The GRDB-free surface the document list and detail views reach for. Each
+/// method resolves the active `CachingBackend` (the production/preview stack) and
+/// forwards to its `(database, serverID)`; without a caching backend (e.g.
+/// `NullRepository` before login) the observations are empty, finished streams
+/// and `fillDocumentQuery` throws.
+extension DocumentStore {
+  /// Eager full-fill of a list: await page 1 + count, then background-page the
+  /// rest. The returned handle carries the `QueryKey` the list then observes.
+  public func fillDocumentQuery(filter: FilterState) async throws -> QueryFillHandle {
+    guard let backend = repository as? any CachingBackend else {
+      throw CachingRepositoryError.cacheMiss
+    }
+    return try await backend.fillQuery(filter: filter)
+  }
+
+  /// Live growing-prefix of a cached query's ordered answer (the list's data).
+  public func observeDocumentPrefix(
+    queryKey: QueryKey, limit: Int
+  ) -> AsyncThrowingStream<[Document], Error> {
+    guard let backend = repository as? any CachingBackend else { return Self.emptyStream() }
+    return backend.database.observeDocumentPrefix(
+      queryKey: queryKey, serverID: backend.serverID, limit: limit)
+  }
+
+  /// Live status of a cached query (scrollbar count, order-stale flag).
+  public func observeQueryStatus(
+    queryKey: QueryKey
+  ) -> AsyncThrowingStream<QueryStatus, Error> {
+    guard let backend = repository as? any CachingBackend else { return Self.emptyStream() }
+    return backend.database.observeQueryStatus(queryKey: queryKey, serverID: backend.serverID)
+  }
+
+  /// Live single document by id, for detail/preview surfaces that must repaint on
+  /// mutation/sync.
+  public func observeDocument(id: UInt) -> AsyncThrowingStream<Document?, Error> {
+    guard let backend = repository as? any CachingBackend else { return Self.emptyStream() }
+    return backend.database.observeDocument(serverID: backend.serverID, id: id)
+  }
+
+  private static func emptyStream<T: Sendable>() -> AsyncThrowingStream<T, Error> {
+    AsyncThrowingStream { $0.finish() }
   }
 }
 
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -186,13 +186,15 @@ public final class DocumentStore: Sendable {
     taskUpdateTask = Task(operation: taskPoller)
   }
 
-  public func clear() {
+  /// Drop the per-server state this store still holds in memory, on a repository
+  /// swap. That is only the task list and the last sync error now: documents
+  /// aren't held in memory under source-of-truth (the list observes the DB
+  /// directly), and the element projection is owned by `elementStore` and
+  /// re-pointed by `wireElementStore()`. `private` because `set(repository:)` is
+  /// the one caller — a swap is the only moment this is the right thing to do.
+  private func clear() {
     tasks = []
     lastSyncError = nil
-    // Documents are not held in memory under source-of-truth — the list observes
-    // the DB directly (see the document-cache surface below). The element
-    // projection is owned by `elementStore` and (re)wired by `wireElementStore()`
-    // on repository change — nothing to clear here.
   }
 
   public func set(repository: some Repository, reload: Bool = true) {
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -739,6 +739,28 @@ extension DocumentStore {
   private static func emptyStream<T: Sendable>() -> AsyncThrowingStream<T, Error> {
     AsyncThrowingStream { $0.finish() }
   }
+
+  /// Debug / maintenance: wipe *all* locally cached data — the GRDB element +
+  /// document caches, the downloaded PDF/thumbnail blobs, the in-memory and
+  /// on-disk image caches, and the per-server delta watermarks — while keeping
+  /// the configured server connections. The live observations repaint empty
+  /// immediately; the next sync / list open refills from the network.
+  public func wipeLocalCache() throws {
+    if let backend = repository as? any CachingBackend {
+      try backend.database.clearCache()
+    }
+    // Downloaded originals/archives/thumbnails (app-group blob store). Rooted at
+    // the app group, so a fresh handle addresses the same files the repository
+    // wrote — no need to reach into the active repository.
+    if let contentStore = try? ContentStore() {
+      try? contentStore.purge()
+    }
+    // Nuke memory + disk image cache.
+    imagePipeline.cache.removeAll()
+    // Per-server changed-metadata delta watermarks (regenerable sync state):
+    // clear them so the reconcile re-baselines over the now-empty cache.
+    DocumentDeltaWatermark.clearAll()
+  }
 }
 
 //// Permissions checking for resources
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -721,7 +721,9 @@ extension DocumentStore {
     }
     lastDocumentReconcile = Date()
     do {
+      // Deletes first (correctness), then the changed-metadata delta (freshness).
       try await backend.reconcileDocumentDeletions()
+      try await backend.reconcileDocumentChanges()
     } catch {
       Logger.shared.info("Document reconcile failed (suppressed): \(error)")
     }
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -703,7 +703,7 @@ extension DocumentStore {
       queryKey: queryKey, serverID: backend.serverID, limit: limit)
   }
 
-  /// Live status of a cached query (scrollbar count, order-stale flag).
+  /// Live status of a cached query (server total, order-stale flag).
   public func observeQueryStatus(
     queryKey: QueryKey
   ) -> AsyncThrowingStream<QueryStatus, Error> {
--- a/AppShared/Sources/AppShared/Networking/CachingRepository.swift
+++ b/AppShared/Sources/AppShared/Networking/CachingRepository.swift
@@ -48,6 +48,13 @@ public protocol CachingBackend: AnyObject, Sendable {
   /// falls back to whatever is already cached).
   func fillQuery(filter: FilterState) async throws -> QueryFillHandle
 
+  /// Remote-delete reconcile (R2): fetch the server's authoritative live id set
+  /// and drop every cached document absent from it — the FK cascade removes them
+  /// from every cached query_order. No-op when nothing is cached. Paperless has
+  /// no deletion feed, so this periodic sweep is how deletes (and trashings)
+  /// disappear locally.
+  func reconcileDocumentDeletions() async throws
+
   /// The shared database and the active server this repository caches into.
   /// `DocumentStore` reads these to point its `ElementStore` projection at the
   /// same `(database, serverID)` the writes land in, so the live observation
@@ -470,6 +477,25 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
 
   public func documents(filter: FilterState) throws -> Wrapped.Documents {
     try wrapped.documents(filter: filter)
+  }
+
+  public func documentIDs(filter: FilterState) async throws -> [UInt] {
+    try await wrapped.documentIDs(filter: filter)
+  }
+
+  public func reconcileDocumentDeletions() async throws {
+    let localIDs = try database.allDocumentIDs(serverID: serverID)
+    // Nothing cached yet → nothing to reconcile (skip the id fetch entirely).
+    guard !localIDs.isEmpty else { return }
+
+    // The unfiltered list is the complete live id set for the server.
+    let serverIDs = Set(try await wrapped.documentIDs(filter: .empty))
+    let removed = localIDs.subtracting(serverIDs)
+    guard !removed.isEmpty else { return }
+
+    Logger.shared.info(
+      "Reconcile: dropping \(removed.count, privacy: .public) remotely-deleted documents")
+    try database.deleteDocuments(serverID: serverID, removedIDs: Array(removed))
   }
 
   public func nextAsn() async throws -> UInt {
--- a/AppShared/Sources/AppShared/Networking/CachingRepository.swift
+++ b/AppShared/Sources/AppShared/Networking/CachingRepository.swift
@@ -594,14 +594,19 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
     try await wrapped.nextAsn()
   }
 
+  /// The version id a document's file-metadata caches under. Both fallbacks
+  /// land on the document id, which equals the root version id server-side: the
+  /// cached row may be absent (nothing fetched it yet) and the read itself may
+  /// fail. In practice the detail view fetches the document first, so the
+  /// versions are usually known by the time this is called.
+  private func fileMetadataVersionID(documentId: UInt) -> UInt {
+    (try? database.document(serverID: serverID, id: documentId))?.currentVersionID ?? documentId
+  }
+
   public func metadata(documentId: UInt) async throws -> Metadata {
     // File-metadata is immutable per file version, so it caches under the
-    // document's current version id (fallback: the document id, which equals the
-    // root version id server-side). The detail view fetches the document first,
-    // so the cached row's versions are usually known by the time we get here.
-    let versionID =
-      (try? database.document(serverID: serverID, id: documentId))?.currentVersionID
-      ?? documentId
+    // document's current version id.
+    let versionID = fileMetadataVersionID(documentId: documentId)
     do {
       let fetched = try await wrapped.metadata(documentId: documentId)
       try database.setFileMetadata(fetched, serverID: serverID, versionID: versionID)
--- a/AppShared/Sources/AppShared/Networking/CachingRepository.swift
+++ b/AppShared/Sources/AppShared/Networking/CachingRepository.swift
@@ -595,21 +595,56 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
   }
 
   public func metadata(documentId: UInt) async throws -> Metadata {
-    try await wrapped.metadata(documentId: documentId)
+    // File-metadata is immutable per file version, so it caches under the
+    // document's current version id (fallback: the document id, which equals the
+    // root version id server-side). The detail view fetches the document first,
+    // so the cached row's versions are usually known by the time we get here.
+    let versionID =
+      (try? database.document(serverID: serverID, id: documentId))?.currentVersionID
+      ?? documentId
+    do {
+      let fetched = try await wrapped.metadata(documentId: documentId)
+      try database.setFileMetadata(fetched, serverID: serverID, versionID: versionID)
+      return fetched
+    } catch {
+      if let cached = try database.fileMetadata(serverID: serverID, versionID: versionID) {
+        Logger.shared.info("metadata(documentId:) network failed (\(error)); serving cached")
+        return cached
+      }
+      throw error
+    }
   }
 
   public func notes(documentId: UInt) async throws -> [Document.Note] {
-    try await wrapped.notes(documentId: documentId)
+    do {
+      let fetched = try await wrapped.notes(documentId: documentId)
+      try database.setNotes(fetched, serverID: serverID, documentID: documentId)
+      return fetched
+    } catch {
+      // `nil` (never cached) is distinct from `[]` (cached, no notes): only the
+      // former propagates the network error.
+      if let cached = try database.notes(serverID: serverID, documentID: documentId) {
+        Logger.shared.info("notes(documentId:) network failed (\(error)); serving cached")
+        return cached
+      }
+      throw error
+    }
   }
 
   public func createNote(documentId: UInt, note: ProtoDocument.Note) async throws
     -> [Document.Note]
   {
-    try await wrapped.createNote(documentId: documentId, note: note)
+    // Pessimistic: the server returns the updated full list, which we write
+    // through so the cached notes stay consistent without a re-fetch.
+    let updated = try await wrapped.createNote(documentId: documentId, note: note)
+    try database.setNotes(updated, serverID: serverID, documentID: documentId)
+    return updated
   }
 
   public func deleteNote(id: UInt, documentId: UInt) async throws -> [Document.Note] {
-    try await wrapped.deleteNote(id: id, documentId: documentId)
+    let updated = try await wrapped.deleteNote(id: id, documentId: documentId)
+    try database.setNotes(updated, serverID: serverID, documentID: documentId)
+    return updated
   }
 
   public func shareLinks(documentId: UInt) async throws -> [DataModel.ShareLink] {
--- a/AppShared/Sources/AppShared/Networking/CachingRepository.swift
+++ b/AppShared/Sources/AppShared/Networking/CachingRepository.swift
@@ -184,7 +184,7 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
     let source = try wrapped.documents(filter: filter)
     let pageSize = Endpoint.defaultDocumentPageSize
 
-    // Page 1 awaited: first window on screen + exact scrollbar count.
+    // Page 1 awaited: first window on screen + the exact total for the count pill.
     let firstPage = try await source.fetch(limit: pageSize)
     let total = await source.totalCount
     try database.writeQueryPage(
--- a/AppShared/Sources/AppShared/Networking/CachingRepository.swift
+++ b/AppShared/Sources/AppShared/Networking/CachingRepository.swift
@@ -409,14 +409,24 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
     try database.deleteElement(SavedViewRecord.self, serverID: serverID, id: savedView.id)
   }
 
-  // MARK: - Documents (forwarded)
+  // MARK: - Documents (pessimistic write-through + cache fallback)
 
   public func update(document: Document) async throws -> Document {
-    try await wrapped.update(document: document)
+    let updated = try await wrapped.update(document: document)
+    // Write the confirmed metadata through; the join observation repaints the row
+    // in place. Written at `.metadata` so the non-downgrade guard preserves an
+    // existing Tier-2 row's detail/permissions (a metadata edit doesn't change
+    // them; permissions edits reconcile on the next detail fetch). Ordering under
+    // the active sort isn't recomputed offline — mark affected queries stale.
+    try database.upsertDocument(updated, serverID: serverID, projectionLevel: .metadata)
+    try database.markQueriesOrderStale(containing: updated.id, serverID: serverID)
+    return updated
   }
 
   public func delete(document: Document) async throws {
     try await wrapped.delete(document: document)
+    // FK cascade removes it from every cached query_order.
+    try database.deleteDocuments(serverID: serverID, removedIDs: [document.id])
   }
 
   public func create(document: ProtoDocument, file: URL, filename: String) async throws {
@@ -424,11 +434,38 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
   }
 
   public func document(id: UInt) async throws -> Document? {
-    try await wrapped.document(id: id)
+    do {
+      guard let fetched = try await wrapped.document(id: id) else {
+        // Gone on the server — drop from cache (cascade clears its query_order).
+        try database.deleteDocuments(serverID: serverID, removedIDs: [id])
+        return nil
+      }
+      // A full-detail fetch — upgrade the row to Tier-2.
+      try database.upsertDocument(fetched, serverID: serverID, projectionLevel: .detail)
+      return fetched
+    } catch {
+      // Offline/transient: serve the last-known cached row (Tier-1 or Tier-2)
+      // rather than failing the open. Mirrors the element offline-first policy.
+      if let cached = try database.document(serverID: serverID, id: id) {
+        Logger.shared.info("document(id:) network failed (\(error)); serving cached")
+        return cached
+      }
+      throw error
+    }
   }
 
   public func document(asn: UInt) async throws -> Document? {
-    try await wrapped.document(asn: asn)
+    do {
+      guard let fetched = try await wrapped.document(asn: asn) else { return nil }
+      try database.upsertDocument(fetched, serverID: serverID, projectionLevel: .detail)
+      return fetched
+    } catch {
+      if let cached = try database.document(serverID: serverID, asn: asn) {
+        Logger.shared.info("document(asn:) network failed (\(error)); serving cached")
+        return cached
+      }
+      throw error
+    }
   }
 
   public func documents(filter: FilterState) throws -> Wrapped.Documents {
--- a/AppShared/Sources/AppShared/Networking/CachingRepository.swift
+++ b/AppShared/Sources/AppShared/Networking/CachingRepository.swift
@@ -40,6 +40,14 @@ public protocol CachingBackend: AnyObject, Sendable {
   /// resource the user lacks permission for is skipped, not fatal.
   func syncElements() async throws
 
+  /// Eager full-fill of a document list (Stage 8 v1): await page 1 (so the first
+  /// window + an exact count land synchronously), write it as the query's order,
+  /// then background-page the rest of the query to the cache. The returned
+  /// ``QueryFillHandle`` carries the `QueryKey` the list observes and a cancel
+  /// handle for the in-flight fill. Throws if page 1 fails (offline → the list
+  /// falls back to whatever is already cached).
+  func fillQuery(filter: FilterState) async throws -> QueryFillHandle
+
   /// The shared database and the active server this repository caches into.
   /// `DocumentStore` reads these to point its `ElementStore` projection at the
   /// same `(database, serverID)` the writes land in, so the live observation
@@ -136,6 +144,44 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
 
       for try await _ in group {}
     }
+  }
+
+  public func fillQuery(filter: FilterState) async throws -> QueryFillHandle {
+    let key = QueryKey(serverID: serverID, filter: filter)
+    let source = try wrapped.documents(filter: filter)
+    let pageSize = Endpoint.defaultDocumentPageSize
+
+    // Page 1 awaited: first window on screen + exact scrollbar count.
+    let firstPage = try await source.fetch(limit: pageSize)
+    let total = await source.totalCount
+    try database.writeQueryPage(
+      queryKey: key, serverID: serverID, documents: firstPage,
+      startPosition: 0, totalCount: total, replaceAll: true, projectionLevel: .metadata)
+
+    // Background-page the rest to disk (append). When this completes the whole
+    // view is local; scrolling then needs no network (v1).
+    let database = database
+    let serverID = serverID
+    let firstCount = firstPage.count
+    let task = Task.detached(priority: .utility) {
+      var position = firstCount
+      do {
+        while !Task.isCancelled {
+          if await source.isExhausted { break }
+          let batch = try await source.fetch(limit: pageSize)
+          if batch.isEmpty { break }
+          try database.writeQueryPage(
+            queryKey: key, serverID: serverID, documents: batch,
+            startPosition: position, totalCount: await source.totalCount,
+            replaceAll: false, projectionLevel: .metadata)
+          position += batch.count
+        }
+      } catch is CancellationError {
+      } catch {
+        Logger.shared.error("Background query fill failed: \(error)")
+      }
+    }
+    return QueryFillHandle(queryKey: key, totalCount: total, fillTask: task)
   }
 
   private func syncCollection<R: ElementRecord>(
--- a/AppShared/Sources/AppShared/Networking/CachingRepository.swift
+++ b/AppShared/Sources/AppShared/Networking/CachingRepository.swift
@@ -55,6 +55,11 @@ public protocol CachingBackend: AnyObject, Sendable {
   /// disappear locally.
   func reconcileDocumentDeletions() async throws
 
+  /// Changed-metadata delta (R3δ): page `ordering=-modified` until older than the
+  /// per-server watermark and refresh the cached rows that changed. Keeps
+  /// already-cached documents fresh without re-opening their list.
+  func reconcileDocumentChanges() async throws
+
   /// The shared database and the active server this repository caches into.
   /// `DocumentStore` reads these to point its `ElementStore` projection at the
   /// same `(database, serverID)` the writes land in, so the live observation
@@ -496,6 +501,72 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
     Logger.shared.info(
       "Reconcile: dropping \(removed.count, privacy: .public) remotely-deleted documents")
     try database.deleteDocuments(serverID: serverID, removedIDs: Array(removed))
+  }
+
+  // The number of changed documents one delta pass will apply before stopping
+  // (a runaway guard; the next pass continues from the advanced watermark).
+  private let deltaCap = 1000
+
+  public func reconcileDocumentChanges() async throws {
+    // Delta only refreshes already-cached metadata; new docs surface via list
+    // fills, not here. Nothing cached ⇒ nothing to refresh.
+    let localIDs = try database.allDocumentIDs(serverID: serverID)
+    guard !localIDs.isEmpty else { return }
+
+    var filter = FilterState.empty
+    filter.sortField = .modified
+    filter.sortOrder = .descending
+    let source = try wrapped.documents(filter: filter)
+
+    guard let watermark = deltaWatermark() else {
+      // First run: establish the baseline from the newest doc; subsequent passes
+      // delta against it. (Avoids re-paging the whole library on cold start.)
+      if let newest = try await source.fetch(limit: 1).first?.modified {
+        setDeltaWatermark(newest)
+      }
+      return
+    }
+
+    var changed: [Document] = []
+    var advanced = watermark
+    pageLoop: while changed.count < deltaCap {
+      let batch = try await source.fetch(limit: Endpoint.defaultDocumentPageSize)
+      if batch.isEmpty { break }
+      for document in batch {
+        guard let modified = document.modified else { continue }
+        // Sorted newest-first: once we reach the watermark, the rest is known.
+        if modified <= watermark { break pageLoop }
+        changed.append(document)
+        if modified > advanced { advanced = modified }
+      }
+      if await source.isExhausted { break }
+    }
+
+    let tracked = changed.filter { localIDs.contains($0.id) }
+    if !tracked.isEmpty {
+      Logger.shared.info(
+        "Reconcile: refreshing \(tracked.count, privacy: .public) changed documents")
+      try database.upsertDocuments(tracked, serverID: serverID, projectionLevel: .metadata)
+    }
+    if advanced > watermark {
+      setDeltaWatermark(advanced)
+    }
+  }
+
+  // Per-server delta watermark (newest `modified` applied). Kept in the app-group
+  // UserDefaults rather than the DB: it is regenerable sync state, not view
+  // state — losing it just re-baselines on the next pass.
+  private var deltaWatermarkKey: String { "documentDeltaWatermark.\(serverID.uuidString)" }
+
+  private func deltaWatermark() -> Date? {
+    let defaults = UserDefaults(suiteName: ContentStore.appGroup) ?? .standard
+    let stamp = defaults.double(forKey: deltaWatermarkKey)
+    return stamp > 0 ? Date(timeIntervalSinceReferenceDate: stamp) : nil
+  }
+
+  private func setDeltaWatermark(_ date: Date) {
+    let defaults = UserDefaults(suiteName: ContentStore.appGroup) ?? .standard
+    defaults.set(date.timeIntervalSinceReferenceDate, forKey: deltaWatermarkKey)
   }
 
   public func nextAsn() async throws -> UInt {
--- a/AppShared/Sources/AppShared/Networking/CachingRepository.swift
+++ b/AppShared/Sources/AppShared/Networking/CachingRepository.swift
@@ -29,6 +29,27 @@ import Persistence
 import SwiftUI
 import os
 
+/// Storage and clearing of the per-server changed-metadata delta watermark.
+///
+/// Non-generic so the keys (and the bulk clear used by "clear local storage")
+/// don't depend on `CachingRepository`'s `Wrapped` type — and so a `static`
+/// constant is even legal (it isn't on the generic type itself).
+enum DocumentDeltaWatermark {
+  static let keyPrefix = "documentDeltaWatermark."
+
+  static func key(serverID: UUID) -> String { "\(keyPrefix)\(serverID.uuidString)" }
+
+  /// Drop every server's watermark so the next reconcile re-baselines. Paired
+  /// with a cache wipe (a stale watermark over an empty cache would skip
+  /// re-fetching the rows the wipe removed).
+  static func clearAll() {
+    let defaults = UserDefaults(suiteName: ContentStore.appGroup) ?? .standard
+    for key in defaults.dictionaryRepresentation().keys where key.hasPrefix(keyPrefix) {
+      defaults.removeObject(forKey: key)
+    }
+  }
+}
+
 /// The cache control surface the store reaches for, kept off the `Repository`
 /// protocol (which stays technology-agnostic). A repository that isn't a
 /// `CachingBackend` (preview, Share Extension, tests) makes the store fall back
@@ -556,7 +577,7 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
   // Per-server delta watermark (newest `modified` applied). Kept in the app-group
   // UserDefaults rather than the DB: it is regenerable sync state, not view
   // state — losing it just re-baselines on the next pass.
-  private var deltaWatermarkKey: String { "documentDeltaWatermark.\(serverID.uuidString)" }
+  private var deltaWatermarkKey: String { DocumentDeltaWatermark.key(serverID: serverID) }
 
   private func deltaWatermark() -> Date? {
     let defaults = UserDefaults(suiteName: ContentStore.appGroup) ?? .standard
--- a/AppShared/Sources/AppShared/Networking/QueryFillHandle.swift
+++ b/AppShared/Sources/AppShared/Networking/QueryFillHandle.swift
@@ -1,0 +1,38 @@
+//
+//  QueryFillHandle.swift
+//  AppShared
+//
+//  Opaque result of kicking off a document-list fill (`CachingBackend.fillQuery`).
+//  Carries the `QueryKey` the list observes, the server-reported total for an
+//  exact scrollbar, and a handle to the background paging task so a refresh /
+//  teardown can cancel the in-flight fill. No GRDB crosses this boundary.
+//
+
+import Foundation
+import Persistence
+
+public struct QueryFillHandle: Sendable {
+  /// The key the list view-model subscribes `observeDocumentPrefix` to.
+  public let queryKey: QueryKey
+  /// Server-reported total from page 1 (exact scrollbar extent), if known.
+  public let totalCount: UInt?
+
+  private let fillTask: Task<Void, Never>
+
+  public init(queryKey: QueryKey, totalCount: UInt?, fillTask: Task<Void, Never>) {
+    self.queryKey = queryKey
+    self.totalCount = totalCount
+    self.fillTask = fillTask
+  }
+
+  /// Cancel the background page-the-rest fill (e.g. on filter change / teardown).
+  public func cancel() {
+    fillTask.cancel()
+  }
+
+  /// Await the background fill completing — primarily for tests and callers that
+  /// want the whole view local before proceeding.
+  public func awaitCompletion() async {
+    await fillTask.value
+  }
+}
--- a/AppShared/Sources/AppShared/Networking/QueryFillHandle.swift
+++ b/AppShared/Sources/AppShared/Networking/QueryFillHandle.swift
@@ -3,8 +3,8 @@
 //  AppShared
 //
 //  Opaque result of kicking off a document-list fill (`CachingBackend.fillQuery`).
-//  Carries the `QueryKey` the list observes, the server-reported total for an
-//  exact scrollbar, and a handle to the background paging task so a refresh /
+//  Carries the `QueryKey` the list observes, the server-reported total behind
+//  the count pill, and a handle to the background paging task so a refresh /
 //  teardown can cancel the in-flight fill. No GRDB crosses this boundary.
 //
 
@@ -14,7 +14,7 @@ import Persistence
 public struct QueryFillHandle: Sendable {
   /// The key the list view-model subscribes `observeDocumentPrefix` to.
   public let queryKey: QueryKey
-  /// Server-reported total from page 1 (exact scrollbar extent), if known.
+  /// Server-reported total from page 1 (the count pill's number), if known.
   public let totalCount: UInt?
 
   private let fillTask: Task<Void, Never>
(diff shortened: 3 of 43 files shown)


paulgessinger:
🤖 _Posted by Claude on Paul's behalf. Backend claims below were verified by reading the paperless-ngx source at `v3.0.0-beta.rc2`; the client-side reading is from this branch. Nothing here was reproduced against a live server — see the caveat at the end._

## R3δ cannot see document-version changes

Recording a gap found while testing the versioning work (#592) interactively, since it lands on this PR's delta rather than that one.

**paperless-ngx does not bump the root document's `modified` when its version set changes.** In all three handlers only the *child* version row is written:

- **Add version** — the view (`views.py:1928-1988`) just enqueues `consume_file`; the consumer builds a new child row and saves *that* (`consumer.py:256-295`, `:589-622`, `:729`). The root is fetched only to build from.
- **Delete version** (`views.py:2037-2093`) — `version_doc.delete()`; `root_doc` is never `.save()`d.
- **Update version label** (`views.py:2123-2170`) — `version_doc.save(update_fields=["version_label"])`, which excludes `modified`, so not even the *child's* timestamp moves.

All three fire `document_updated` with the root, but its receivers only read `modified`, push a websocket frame, or enqueue LLM indexing. The one path that would bump it is a user-configured "Document Updated" workflow (`handlers.py:986-996`) — opt-in, and not something to rely on.

**Why that matters here.** `reconcileDocumentChanges()` (`CachingRepository.swift:696-756`) pages `ordering=-modified` and stops at the first row at or below the stored `deltaWatermark`:

```swift
guard let modified = document.modified else { continue }
if modified <= watermark { break pageLoop }
```

A version add/delete leaves `modified` untouched, so the changed document never rises above the watermark and R3δ walks straight past it. **No amount of waiting fixes it** — the passive freshness guarantee simply does not cover version-set changes.

Concretely, with version-keyed content (#592) the stale row yields a stale `currentVersionID`, hence a stale `?version=` on both the thumbnail URL and the download key.

### What still works, so this is bounded

- **Pull-to-refresh on the list** → `fillQuery` (`CachingRepository.swift:247-288`) re-fetches unconditionally, no `modified` comparison, so it *does* fix the cached row (page 1 immediately, the rest via the background continuation).
- **Opening the detail view** → `document(id:)` write-through fixes that row.

So it is "invisible to passive sync" rather than "unfixable" — but the app's stated model is that the cache converges on its own, and for this class of change it does not.

### Options

1. **Accept and document** — note in `reconcileDocumentChanges()` that version-set changes are outside R3δ's detection, so nobody later assumes the delta is complete. Cheapest, and honest.
2. **Report upstream.** A version add/delete *is* a change to the document as an API client observes it (`versions` is a serialized field, and it changes what `/download/` and `/thumb/` return by default). Not bumping `modified` looks like a genuine backend bug, and it breaks any timestamp-based sync client, not just this one. Fixing it there makes R3δ correct for free.
3. **Widen the delta** — no cheap version of this. There is no server-side signal to key on short of re-fetching documents wholesale, which is what R3δ exists to avoid.

My suggestion is 1 + 2: a comment here now, and an upstream issue, rather than contorting the delta around a backend behaviour that is likely to be considered a bug.

### Caveat

Verified by source reading only. I have not confirmed against a running v3.0.0-beta instance that `modified` is unchanged after a version delete, and that check is worth doing before filing upstream — a single `GET /api/documents/{id}/` before and after would settle it.

paulgessinger:

## Closing out the R3δ note above

Re-checked this while reviewing the PR. It has been overtaken by events, but only partly.

**The upstream half worked.** [paperless-ngx#13170](https://github.com/paperless-ngx/paperless-ngx/pull/13170) — *"Change: update root modified timestamp on version changes"* — merged **2026-07-20**, the day after the comment above was posted. The root document's `modified` now bumps on version add/delete/label change, so R3δ sees version-set changes on any server at or past that beta. Option 2 is done.

**The in-code half was never done.** Option 1 was to note the limitation in `reconcileDocumentChanges()` so nobody later assumes the delta is complete. There is no such comment — I checked at this PR's head and at the stack tip. The doc comment at `CachingRepository.swift:79-81` still describes the delta without the caveat.

**The residual gap is real but bounded**, and is already recorded as item 12 in the review ledger (#656): servers older than that beta stay blind to version-only changes in the background sweep. The UI doesn't depend on it — `DocumentCell`/`DocumentPreview` key their `.task` on `versionQueryID`, and `DocumentDetailModel` compares `currentVersionID` before `modified` — so the gap is confined to the sweep. No action planned there.

So the only thing outstanding from the comment above is a two-line caveat comment. Worth folding into whatever change addresses **H3** in the review below, since that touches the same function and the same doc comment.

paulgessinger:

## Review pass on #600

**Method and caveats up front.** Everything below comes from reading this PR's code **at its head** (`feat/document-cache-source-of-truth`, `99d8028`) via `jj file show`, not the working copy — the stack tip is 45 commits further on and shows different code. Three parallel passes (Persistence, caching/networking, app/UI), deduplicated and spot-verified by hand.

- **Executed:** only `swift test --package-path Persistence` (93/93 pass) against #600 + a trial fix for H1.
- **Not executed:** nothing was driven in the simulator. No finding below was reproduced against a live server.
- **Inferred, not verified:** M5 rests on Postgres pagination behaviour under a non-unique ordering, which I could not check from this repo.

I also **verified that all four high-severity findings survive unchanged to the stack tip** — none of them is fixed by a later PR.

---

## High

### H1 · A cached list can contain the same document twice — ✅ fixed

`query_order`'s primary key is `(server_id, query_key, position)` (`V4_CreateDocumentCache.swift:64`) with no uniqueness on `remote_id`, `writeQueryPage` (`Database+Documents.swift:57-63`) appends each page with no check for an id already placed under the key, and the read path `queryWindowSQL` (`Database+Documents.swift:159-166`) is a plain `JOIN … ORDER BY q.position` with no `DISTINCT`.

This is a **regression**. `main`'s `DocumentListViewModel.swift:42-46` carried a `seenIds: Set<UInt>` dedup whose comment names precisely this failure:

> The server can return the same document on adjacent pages if a new doc is inserted between fetches and shifts the page offsets — without dedup this surfaces as a SwiftUI ForEach duplicate-ID warning and stutters.

That dedup was deleted in this PR and nothing replaced it, at #600 or anywhere up to the stack tip.

**Failure:** >250 documents, background fill in flight, a document is added (another client, the consumer, the share extension) between page 1 and page 2. The document is returned on both pages and written at two positions. `DocumentList.swift:259` renders `ForEach(…, id: \.1.id)` → duplicate-ID runtime warning, a visibly duplicated row, unstable diffing, and broken `.id(document.id)` scroll targeting. Unlike the old in-memory dedup this is now **persisted**, so it survives until the next `replaceAll` fill.

Secondary effect: `fetchQueryStatus` derives `localCount` from a `fetchCount` of `query_order` rows (`Database+Documents.swift:174-178`), so a duplicate also inflates prefix-completeness accounting — `localCount` can reach `totalCount` while the cache holds fewer distinct documents than the query has.

### H2 · An interrupted fill leaves a permanently truncated list, with no error and no resume — ➡️ deferred to #618

Two halves, both required:

- **Producer** — `CachingRepository.swift:190-192` writes page 1 with `replaceAll: true`, destroying the key's entire existing `query_order` *before* the background continuation has fetched anything. Both of the continuation's failure modes are silent: an empty `catch is CancellationError {}` and a bare `Logger.shared.error` (`:212-215`). `writeQueryPage` has already stamped `query_meta` with the server's full `totalCount` and `orderStale: false`, so a truncated cache is indistinguishable from a complete one.
- **Consumer** — `DocumentListViewModel.fetchMoreIfNeeded` (`:194-201`) only widens the observed prefix. There is no network path left (`// Pure local windowing … No network`). Nothing on scroll, foreground, or `sync()` restarts the fill.

**Failure:** 3,000-document list fully cached. Pull to refresh. Page 1 (250) succeeds and wipes positions 250–2999; connectivity drops during page 2. The count pill still says 3,000; scrolling stops dead at 250 with no spinner and no error. `documents.count < prefixLimit` also means the widen guard can never fire again. Recovery requires another *fully successful* fill. Backgrounding the app mid-fill (iOS suspends the detached task) does the same thing.

Net: **a refresh can leave the offline cache strictly worse than before it ran** — which is the one thing an offline feature must not do.

### H3 · R3δ's runaway cap advances the watermark to the newest change, permanently discarding everything below it — ✅ fixed (on #617)

`CachingRepository.swift:551-574`. `advanced` is seeded from the watermark and only ever raised (`:561`), and the walk is `ordering=-modified` (newest first) — so after the first batch `advanced` already equals the newest document in the library. When the loop exits because `changed.count < deltaCap` went false, rather than by reaching the watermark or exhausting the source, `:572-573` still commits that newest value.

The comment at `:528` — *"a runaway guard; the next pass continues from the advanced watermark"* — describes behaviour a single high-water mark structurally cannot produce. It cannot express "applied `[oldest, newest]` but not `[old_watermark, oldest)`".

**Failure:** a server-side bulk retag touches 5,000 documents. One pass collects ~1,000, refreshes those, and sets the watermark to the newest of them. The remaining ~3,800 cached documents keep their stale tags/titles/correspondents **forever** — no later pass revisits them short of a cache wipe. Also reachable just by not opening the app for a long time.

This gets worse upstack, where *Entire library* mode leans on this same delta to keep the library current between full fills.

**Sketch:** track `oldestApplied` too; set the watermark to `advanced` only when the walk terminated by reaching the watermark or exhausting the source, and to `oldestApplied` when the cap was hit.

### H4 · Both list observations retain `self` for the process lifetime — ✅ fixed

`DocumentListViewModel.swift:217-218` and `:233-234`:

```swift
documentTask = Task { @MainActor [weak self] in
  guard let self else { return }        // ← strong, held across every suspension
  for try await docs in store.observeDocumentPrefix(…) { … }
```

`guard let self` promotes to a strong local that lives for the whole infinite `for await`, and the task is stored on `self` — so `self → documentTask → frame → self`. `deinit` (`:80-84`) can never run. `ElementStore` gets this right (`ElementStore.swift:130`, `:148`, `:165` all re-check *inside* the loop), which makes this look like an oversight rather than intent.

**Failure:** on iPad, rotate between portrait and landscape (or collapse/expand the sidebar) so `DocumentView` swaps `compactBody`/`contentColumn`. SwiftUI releases the old view model but it never deallocates; its two `ValueObservation` streams keep running forever, re-fetching and decoding up to `prefixLimit` `DocumentRecord`s on every DB write and pinning the previous `DocumentStore`. Each rotation stacks another.

**Sketch:** resolve `store` up front, then `guard let self else { break }` inside the loop.

---

## Medium

**Round 2 status.** One fixed on this PR: **M5** (the remote-delete sweep paged a non-unique ordering). **M9 is withdrawn — it was not a bug** (see below). **M6 is withdrawn** — already fixed, by `1a97237c4` on #620 *plus* `10021dc3` on #627; the collector dedups with `==` and bridged `URLError`s carry a per-request task id, so the collapse only works once transport failures are normalized into the `Equatable` `RequestError`. The rest are deferred with owners: **M1** and **M4** → #618, **M2** → #617 (needs a design call — that PR deliberately moved ordering repair into `reconcileSavedViewMembership`), **M7** → #620, **M10** → #627 (folded into ledger item 9, same observation-region problem). **M3** and **M8** have no owner and sit in the ledger: M3 is inert (nothing reads `orderStale`), M8 needs a product decision about whether the debug menu is a shipped feature. All of it is ledger #656, items 19-24.

**M9 was my mistake, and the fix is reverted.** I flagged "nothing re-populates the app after Clear Cached Data" as a defect and shipped a fix that emitted `.repositoryChanged` and kicked a sync. That destroys the button's actual purpose: it exists to leave an **empty database** so cold offline launches can be tested. Auto-refilling means the empty state survives only if you happen to be offline when you tap it — turning a reliable tool into a trap. The commit has been dropped from this PR and the two conflict resolutions it caused upstack were rewound with it.

The one real sub-point survives: the confirmation copy promises "everything re-downloads as needed", which is misleading for a button whose job is to leave things empty. That is a wording fix, not a behaviour one.

This also answers the product question left open in **M8** / ledger item 24: it is a debug tool, not a shipped feature. So it should be `#if DEBUG`-gated, and its strings should be `Text(verbatim:)` rather than the 66 localized lines currently exported to Crowdin — matching the sibling "Export servers to legacy storage" action in the same file.

One thing found while fixing M5 that isn't in the list below: **H1's `.upsert` → `.insert` change made a *position* collision fatal.** `upsert` emitted `ON CONFLICT DO UPDATE` with no target and silently overwrote; a plain insert hits the primary key's default `ABORT`. Confirmed with a test — `SQLite error 19: UNIQUE constraint failed: query_order…position` — and the background continuation swallows it, killing the rest of the fill. Fixed by giving the primary key an explicit `.replace` in V4, mirrored into V6, with a regression test. It matters because M1 shows those collisions are now routine.


1. **No in-flight registry per `QueryKey`** (`CachingRepository.swift:182-218`, `DocumentListViewModel.swift:176-190`). `fill?.cancel()` can only cancel an *already-assigned* handle, and cancellation is re-checked only at the top of the background loop (`:202`) — never between `source.fetch` returning (`:204`) and `writeQueryPage` (`:206`). Two fills on the same key interleave independent position counters, which feeds H1. Reachable without any race at all when two view models' filters hash to the same key. `.repositoryChanged` starts two such fills (`DocumentList.swift:205-213` → `reload()` and `filterState.clear()` → `refresh(filter:)`).
2. **R3δ never marks queries order-stale** (`CachingRepository.swift:566-571`). `update(document:)` deliberately calls `markQueriesOrderStale(containing:)` at `:454` because ordering isn't recomputed offline; the reconcile applies the same class of change and calls nothing. Rename a doc from "Zebra" to "Aardvark" server-side and it renders last in an A→Z list with no staleness signal.
3. **An append silently clears `order_stale`** (`Database+Documents.swift:64-66`). `writeQueryPage` passes `orderStale: false` unconditionally, including on the `replaceAll: false` append path, which does not re-establish positions 0..n. Latent today — nothing reads the flag — but it makes the flag wrong the moment something does.
4. **A 403 is caught as "offline" and the cached document is served anyway** (`CachingRepository.swift:479-487`, `:495-501`, `:609-615`, `:623-631`). The catch is unqualified, so `RequestError.forbidden` takes the same path as a transport failure. An admin revokes view permission; the user still gets the full cached title, notes and PDF, with no indication.
5. **The R2 delete sweep pages a non-unique ordering and drives destructive deletes** (`ApiRepository.swift:555-561`). `FilterState.empty` carries `sortField: .asn`, so the sweep issues `ordering=-archive_serial_number&page_size=25000&fields=id`. ASN is NULL for most documents, so nearly the whole result set is one tie group, paged by offset. A document missed across a page boundary lands in `removed` and is deleted from the cache. Gated on >25,000-document libraries. *This is the one finding I could not fully verify — it depends on Postgres tie behaviour.* Cheap hardening regardless: order the id sweep by `id`.
6. **Detail pull-to-refresh fans one failure into up to four toasts** (`DocumentDetailModel.swift:96-102`). `load(onError:)` hands the same `onError` to `loadDocument`, `loadMetadata`, `loadNotes` and `loadSuggestionsQuietly`, which run concurrently. The doc comment justifies this by `ErrorController.shouldSuppress` — but that short-circuits unless `NetworkMonitor.isOnline == false` (`ErrorSuppression.swift:30`). Server down while on Wi-Fi ⇒ four stacked toasts. The list got first-error coalescing for exactly this; detail did not.
7. **The notes prefetch toasts a permissions error** (`DocumentDetailModel.swift:119-127`). Its own doc comment says it is *"Routed through `store.notes(for:)` so the `.note` view-permission gate is respected"* — but respecting the gate here means `checkPermission` throws `PermissionsError` (`DocumentStore.swift:315`), which is `DisplayableError` and becomes a toast. `DocumentNoteView` guards with `canViewNotes` before loading; this call site doesn't. A restricted user gets "You don't have permission to view notes" on every refresh of every document.
8. **`wipeLocalCache()` blocks the main thread and ships to everyone** (`DocumentStore.swift:748-763`). `DocumentStore` is `@MainActor` and this is synchronous: a blocking `writer.write` `DELETE FROM` across ~15 tables, a recursive `removeItem` over the whole blob directory, plus `imagePipeline.cache.removeAll()`. `DebugMenuView` is reached from `SettingsView.swift:211` in the plain **Advanced** section with no `#if DEBUG`. On a multi-GB cache this is watchdog territory.
9. **After "Clear Cached Data" nothing re-syncs** (`DebugMenuView.swift:49-56`). Connection safety is fine — I verified `clearCache()` cannot touch `server` rows and `ContentStore.purge()` is scoped to `Caches/ContentStore`. But the observations repaint empty and nothing refills: `load()` won't re-run (`guard documentTask == nil`) and `.task` doesn't re-fire on sheet dismissal. The user is told "everything re-downloads as needed" and lands on an empty list, empty sidebar and empty filter sheet until a manual pull-to-refresh.
10. **The prefix observation re-fires on every fill page** (`Database+Observe.swift:65-77`). No `.removeDuplicates()`, and the tracked region is the whole `document` + `query_order` tables, so a 20,000-document fill re-runs the join and re-decodes the 250-row window ~80 times, assigning `documents` each time and invalidating the list body for no visible change.

## Low / latent

`ContentStore.purge()` reports success when the wipe failed (`try?` at `ContentStore.swift:151-153`); a cache-write failure is reported to the user as a failed *server* mutation (`CachingRepository.swift:454-462`) even though the server write already committed; R3δ's first-run baseline strands changes made between the seeding fill and the first pass (`:542-549`); the `modified <= watermark` compare can drop a change inside one millisecond bucket because the decoder truncates to milliseconds while paperless emits microseconds; `ApiRepository.documentIDs` hardcodes `searchApi: .legacy` (`:559`) while its neighbour passes the negotiated value; `DocumentStore.swift:725-726` runs R2 then R3δ under one suppressing catch, so the delta never runs when the (more failure-prone) delete sweep throws; a failed PDF download is never retried by pull-to-refresh (`DocumentDetailModel.swift:159-198`, `.error` falls into `default: break`); the new document-cache accessors bypass `Database.wrapping`, so raw `GRDB.DatabaseError`s escape the package contrary to the invariant stated at `Database+Elements.swift:7-8`; and `V4`'s `server_sync_state.delta_watermark` column is dead at this stage — the watermark actually lives in app-group `UserDefaults`, which puts it outside `clearCache()`'s transaction.

**Test gaps:** `appendPreservesOrder` appends two *disjoint* pages, so H1 is untested; `fillClearsStale` only exercises `replaceAll: true`, so it asserts the correct case and skips the broken one; `DocumentCacheTests` registers exactly one server throughout, so `allDocumentIDs` — the input to a destructive reconcile — has no two-server assertion.

## Not findings (checked, clean)

Migrations (`V4`/`V5` are new files appended to the migrator; `V1`–`V3` untouched — no forward-only violation); **multi-server scoping**, which I probed hard and could not break — every read, write and observation threads `serverID`, `QueryKey` hashes it with a domain separator, and the `queryWindowSQL` join is on `(server_id, id)` not just `id`; localization (all six new `Settings.xcstrings` keys are referenced, no new hardcoded user-facing strings); the empty-state flash guard; pull-to-refresh self-cancellation; `.task` idempotency across push/pop; and the `TagsEditSheet` tag rollback.

Two findings from the first pass turned out to be **already fixed upstack** and are withdrawn: orphaned `document_note`/`file_metadata` rows, and missing document eviction — `pruneUnreferencedDocuments` exists at the stack tip.

---

## Status after this round

Three of the four are fixed; **H2 is the one still open**, and it is the one I would not ship as-is.

| | Where it landed |
|---|---|
| **H1** | #600 — `uniqueKey(…, onConflict: .ignore)` on `query_order` in V4 + `.upsert` → `.insert`, with two tests. Repeated on **#617**, because `V6_DropProjectionAndQueryOrderFK` rebuilds the table and a rebuilt table does not inherit the constraint — verified by the test failing at the stack tip with the V4-only version. `replaceQueryOrder` switched to `.insert` there too. |
| **H3** | **#617**, not here. See the correction below. |
| **H4** | #600 — resolve `store` before the loop, re-check `self` inside it. |
| **H2** | **Deferred to #618** — see the decision below. Recorded as item 18 in the review ledger (#656) and as an explicit item on that PR. |

### Correction: the `oldestApplied` fix I proposed for H3 does not work

The original text suggested advancing the watermark to the oldest change applied when the cap is hit. That is wrong, and wrong in the same direction as the bug.

Walking newest-first, the documents above the watermark are `d1 … dN` descending. The cap stops the walk after `d1 … d1000`, so `oldestApplied = modified(d1000)` and the entire unapplied backlog `d1001 … dN` lies *below* it. Setting the watermark to `modified(d1000)` means the next pass stops there — losing exactly the same range as setting it to `modified(d1)`, plus re-applying the top 1000 for nothing. A high-water mark only moves up, so no scalar value can encode "applied `[oldest, newest]` but not `[watermark, oldest)`" — which is a sentence the finding itself contains. Thanks to @paulgessinger for catching it.

**What actually landed** is a change of walk direction: the catch-up now runs *oldest-first from the watermark*, committing the cursor after each page. `deltaCap` becomes an honest per-pass budget — a pass that stops early (capped, cancelled, or killed on a background time budget) leaves the watermark at the last document applied, and the next pass resumes from there. The server-side bound is date-granular and exclusive (`modified__date__gt`, no `gte`), which `FilterState` already compensates for, so a pass re-fetches from the start of the watermark's day. Documents sharing the cursor's exact timestamp are re-applied rather than skipped, which is idempotent.

One known bound, called out for review rather than engineered around: if more than `deltaCap` documents share a single `modified` timestamp the cursor cannot advance past them and successive passes re-apply the same group. That needs 1000+ rows inside one millisecond (the decoder truncates to milliseconds).

It went on **#617** because that PR's `reconcileDocumentChanges` is a different function — `entireLibrary` gating, `fullPerms`, `.full` projection, `invalidateNotes`. Applying the rewrite here conflicted from #617's first commit onward; applying it there is conflict-free against the finished version. The cost is that this PR ships with R3δ non-convergent, which degrades passive freshness after a large bulk edit but corrects on any list refresh or document open.

### Also corrected: I quoted a rebase-conflict count earlier, and it was measured wrong

The figure came from a revset that walks *ancestors* of the working copy, so it excluded every descendant and measured nothing. I am not replacing it with a new number — the corrected count swung entirely on where a single comment line sat relative to a line #617 rewrites, which is a good argument for not quoting the metric at all. What matters is stated per-fix above.

### Decision on H2: it lands on #618, and this PR ships without it

Two corrections to the finding as originally written. First, the "scrolling can't fetch more" half is **not** a defect — it's deliberate and documented (`plans/local-database-source-of-truth-variant.md`, *"Why v1 stays the shipped behavior"*), with R3b deferred as the proper fix, and the plan explicitly accepts the online ceiling that follows. What that reasoning assumes is that the fill *completes* — "every opened view eager-fills in full". The real defect is that a fill dying partway breaks v1's own promise: destructively (page 1 wipes the previous order first), silently (both catch arms swallow), indistinguishably (`query_meta` carries the full `totalCount` either way), and with no retry.

Second, the fix belongs upstack. `fillQuery` is structurally identical at this PR's head and #618's, and `DocumentListViewModel` diverges by only 11 lines — so the change is the same size in either place. But the half that makes the failure *visible* is impossible here: `query_sync_error` (V7), `recordQuerySyncError` and `DocumentStore.syncErrors` all arrive with #618, where `fillLibrary` already uses them. And #617/#618 edit the exact lines the fix touches — the `NetworkTransfer.$category.withValue(.fill)` closure boundary, the `projectionLevel:` argument, and the `Logger.shared` → `Logger.sync` rename in the catch arm — which is the same collision that made H3 unlandable here. The `filled_at` route needs no migration, so the "V4 is unshipped" deadline that applied to H1 doesn't apply to this one.

**The cost, stated plainly: this PR goes to TestFlight with H2 live.** It is the one finding from this round where using the app makes things worse rather than merely failing to improve them — a pull-to-refresh that loses connectivity mid-fill leaves less cached than before it ran, with no error and nothing that resumes. The recovery is another successful pull-to-refresh, which the user has no reason to know is needed.

Full write-up and the proposed fix are on #618 and in ledger item 18 (#656).

### Verification

App builds clean at the stack tip; Persistence 122/122 there, 95/95 at this PR's head; no conflicted commit anywhere in the stack. Nothing was driven in the simulator. The R2 tie-ordering finding (M5) still rests on Postgres pagination behaviour I could not check from this repo.

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -180,16 +186,13 @@ public final class DocumentStore: Sendable {
     taskUpdateTask = Task(operation: taskPoller)
   }
 
-  public func clearDocuments() {
-    documents = [:]
-  }
-
   public func clear() {
```
paulgessinger: Do we need the function then?

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -658,6 +667,102 @@ public final class DocumentStore: Sendable {
   }
 }
 
+// MARK: - Document cache surface (Stage 8)
```
paulgessinger, line 672: Let's do another pass here and remove references to the plan stages please!

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -29,6 +29,27 @@ import Persistence
...
+  static func key(serverID: UUID) -> String { "\(keyPrefix)\(serverID.uuidString)" }
+
+  /// Drop every server's watermark so the next reconcile re-baselines. Paired
+  /// with a cache wipe (a stale watermark over an empty cache would skip
+  /// re-fetching the rows the wipe removed).
+  static func clearAll() {
+    let defaults = UserDefaults(suiteName: ContentStore.appGroup) ?? .standard
+    for key in defaults.dictionaryRepresentation().keys where key.hasPrefix(keyPrefix) {
+      defaults.removeObject(forKey: key)
+    }
```
paulgessinger, line 49: Is this a common pattern? The point is to catch all watermarks based on the prefix, right?

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -40,6 +61,26 @@ public protocol CachingBackend: AnyObject, Sendable {
...
+  /// Remote-delete reconcile (R2): fetch the server's authoritative live id set
+  /// and drop every cached document absent from it — the FK cascade removes them
+  /// from every cached query_order. No-op when nothing is cached. Paperless has
+  /// no deletion feed, so this periodic sweep is how deletes (and trashings)
+  /// disappear locally.
+  func reconcileDocumentDeletions() async throws
+
+  /// Changed-metadata delta (R3δ): page `ordering=-modified` until older than the
+  /// per-server watermark and refresh the cached rows that changed. Keeps
+  /// already-cached documents fresh without re-opening their list.
```
paulgessinger, line 81: This probably needs to be updated given the changes landed.

paulgessinger:

## Review round closed

Where every finding ended up. The PR description now carries the commit-by-commit list; this is the disposition summary.

**Fixed on this PR (6 commits):** H1 duplicate documents in a cached query, H4 the list view model never deallocating, the position-collision regression H1's own fix introduced, M5 the delete sweep paging a non-unique ordering, and two lows — the id sweep's hardcoded legacy search API, and `ContentStore.purge()` reporting success on a failed wipe.

**Fixed on #617 (3 commits):** H3, the delta stranding any backlog larger than its cap, plus the two `V6`-rebuild carry-overs the #600 schema changes needed.

**Deferred with owners:** H2 and M1/M4 → #618; M2 → #617 (needs a design call); M7 and a PDF-retry low → #620; M10 → #627. All in #656, items 18–26.

**Withdrawn, three of them:**
- **M6** (detail view fanning one failure into four toasts) — already fixed, but only by `1a97237c4` on #620 *and* `10021dc3` on #627 together: the collector dedups with `==`, and bridged `URLError`s carry a per-request task id, so nothing ever compared equal until transport failures became the `Equatable` `RequestError`.
- **M9** (nothing re-populates the app after Clear Cached Data) — **my mistake.** I read an empty app as a defect; the button exists to leave an empty database for testing cold offline launches, so empty *is* the deliverable. I wrote a fix, it broke that workflow, and it has been reverted along with the two upstack conflict resolutions it caused. The residue is only that the confirmation copy promises "everything re-downloads as needed", which is misleading for a button meant to leave things empty.
- Two lows: the millisecond-bucket watermark tie (fixed by H3's strict cursor comparison) and the dead `delta_watermark` column (resolved by #617 moving cursors into the database).

**One product question got answered along the way:** the clear-cache action is a debug tool, not a shipped feature. So it wants `#if DEBUG` gating and `Text(verbatim:)` strings rather than the 66 localized lines currently exported to Crowdin — matching the sibling "Export servers to legacy storage" action in the same file. Recorded as #656 item 24.

### What was and wasn't verified

Every finding came from reading code **at each PR's head**, not the working copy — the stack tip is 40-odd commits further on and shows different code. Builds are clean at the tip, all four packages pass (Persistence 123, Networking 215, DataModel 230, Common 61), and no commit in the stack is conflicted.

**Nothing was driven in the simulator, and nothing was reproduced against a live server.** Two claims rest on reasoning I could not close from this repo: the delete sweep's tie-ordering behaviour depends on Postgres pagination, and the concurrent-fill race's timing window is inferred from the code rather than observed.

I also published two things in this thread that were wrong and have since been corrected in place — a fix proposal for H3 that could not work, and a rebase-conflict count measured with a revset that excluded every commit it was supposed to cover.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -363,52 +442,209 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
...
+    }
   }
 
   public func documents(filter: FilterState) throws -> Wrapped.Documents {
     try wrapped.documents(filter: filter)
   }
 
+  public func documentIDs(filter: FilterState) async throws -> [UInt] {
+    try await wrapped.documentIDs(filter: filter)
+  }
```
paulgessinger, line 510: I this used at all?

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -363,52 +442,209 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
...
+
+  public func reconcileDocumentDeletions() async throws {
+    let localIDs = try database.allDocumentIDs(serverID: serverID)
+    // Nothing cached yet → nothing to reconcile (skip the id fetch entirely).
+    guard !localIDs.isEmpty else { return }
+
+    // The unfiltered list is the complete live id set for the server.
+    let serverIDs = Set(try await wrapped.documentIDs(filter: .empty))
```
paulgessinger, line 518: We've established that `.empty` sorts by `ASN`. I don't remember if that was an intentional choice or not, but maybe we could change the default?

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -363,52 +442,209 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
...
+  private var deltaWatermarkKey: String { DocumentDeltaWatermark.key(serverID: serverID) }
+
+  private func deltaWatermark() -> Date? {
+    let defaults = UserDefaults(suiteName: ContentStore.appGroup) ?? .standard
+    let stamp = defaults.double(forKey: deltaWatermarkKey)
+    return stamp > 0 ? Date(timeIntervalSinceReferenceDate: stamp) : nil
+  }
+
+  private func setDeltaWatermark(_ date: Date) {
+    let defaults = UserDefaults(suiteName: ContentStore.appGroup) ?? .standard
```
paulgessinger, line 589: Centralize defaults so they cannot go out of sync?

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -363,52 +442,209 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
...
 
   public func metadata(documentId: UInt) async throws -> Metadata {
-    try await wrapped.metadata(documentId: documentId)
+    // File-metadata is immutable per file version, so it caches under the
+    // document's current version id (fallback: the document id, which equals the
+    // root version id server-side). The detail view fetches the document first,
+    // so the cached row's versions are usually known by the time we get here.
+    let versionID =
+      (try? database.document(serverID: serverID, id: documentId))?.currentVersionID
+      ?? documentId
```
paulgessinger: That's a lot of optional handling

AppShared/Sources/AppShared/Networking/QueryFillHandle.swift
```diff
@@ -0,0 +1,38 @@
...
+//  teardown can cancel the in-flight fill. No GRDB crosses this boundary.
+//
+
+import Foundation
+import Persistence
+
+public struct QueryFillHandle: Sendable {
+  /// The key the list view-model subscribes `observeDocumentPrefix` to.
+  public let queryKey: QueryKey
+  /// Server-reported total from page 1 (exact scrollbar extent), if known.
```
paulgessinger: We never actually pipe this to the scrollbar.

Networking/Sources/Networking/Api/ApiRepository.swift
```diff
@@ -550,6 +550,27 @@ extension ApiRepository: Repository {
...
+  /// Cheap id-only projection (`fields=id`) in a few very large pages — the
+  /// authoritative ordered id set for the remote-delete reconcile.
+  public func documentIDs(filter: FilterState) async throws -> [UInt] {
+    Logger.networking.notice("Getting document id set for filter")
+    // Page over a *unique* ordering. The caller's sort carries no meaning here
+    // (the result is consumed as a set), and paging a non-unique one is not
+    // stable — `FilterState.empty` sorts by archive serial number, which is NULL
+    // for most documents, so nearly every row lands in one tie group. Any id
+    // dropped across a page boundary reads as a remote deletion to the
+    // reconcile, which then evicts that document from the cache.
```
paulgessinger: Simplify

Networking/Sources/Networking/Api/ApiRepository.swift
```diff
@@ -550,6 +550,27 @@ extension ApiRepository: Repository {
     return ApiPagedSource<ApiDocument, Document>(cursor: cursor, map: { $0.domain })
   }
 
+  /// Cheap id-only projection (`fields=id`) in a few very large pages — the
+  /// authoritative ordered id set for the remote-delete reconcile.
+  public func documentIDs(filter: FilterState) async throws -> [UInt] {
```
paulgessinger, line 555: The `Repository` protocol defaults this. There's no way for the default implementation to reach for the concrete cursor and do what this function here does, right?

Persistence/Sources/Persistence/Migrations/V4_CreateDocumentCache.swift
```diff
@@ -0,0 +1,115 @@
...
+
+/// Document-metadata cache tables (Stage 8).
+///
+/// Two jobs, two tables:
+///
+/// - `document` — one row per `(server_id, id)`, query-independent. A document
+///   that appears in many lists still has a single metadata row here. Indexed
+///   columns carry what the join/lookups need (`title`, `asn`, the completeness
+///   marker `projection_level` / `detail_fetched_at`); the long tail lives in the
+///   `data` JSON blob, exactly like the element cache.
```
paulgessinger, line 11: Are we properly using sqlite's JSON support for these?

Persistence/Sources/Persistence/Migrations/V4_CreateDocumentCache.swift
```diff
@@ -0,0 +1,115 @@
...
+
+    // Regenerable per-server sync cursors (one row per server). Stored as REAL
+    // (`timeIntervalSinceReferenceDate`) for exact round-trips — the delta
+    // comparison is precision-sensitive. Cleared by `clearCache` (see `tables`).
+    try db.create(table: "server_sync_state", options: [.strict]) { t in
+      t.column("server_id", .blob)
+        .notNull()
+        .references("server", onDelete: .cascade)
+      // Newest document `modified` applied by the changed-metadata delta (R3δ).
+      t.column("delta_watermark", .real)
```
paulgessinger: I thought the watermark is in userdefaults?

Persistence/Sources/Persistence/Migrations/V5_CreateDocumentDetailCache.swift
```diff
@@ -0,0 +1,44 @@
...
+///   `/notes/` endpoint cached here.
+/// - `file_metadata` — one row per `(server_id, version_id)`, holding the
+///   `/metadata/` sub-resource (checksums, sizes, mime, parsed item lists) as a
+///   JSON `data` blob. **Immutable per file version** (like `ContentStore`), so
+///   keying by version means a cached copy never goes stale until the version
+///   changes.
+///
+/// Both FK-reference `server(id)` with `ON DELETE CASCADE`, so removing a
+/// connection tears down its detail cache along with the rest.
+enum V5_CreateDocumentDetailCache {
```
paulgessinger, line 21: Should this just be part of V4? Any benefit to a second migration?

Persistence/Sources/Persistence/Models/QueryKey.swift
```diff
@@ -0,0 +1,55 @@
...
+/// Two notes on correctness:
+/// - The hash is **CryptoKit `SHA256`**, not Swift's `Hasher` — `Hasher` is
+///   seeded per process, so a `Hasher`-derived key would change every launch and
+///   never match yesterday's cached rows.
+/// - The canonical input is `FilterRule.queryItems(for:)` (which already sorts
+///   multi-value rule values) re-sorted as `name=value` lines, so it is
+///   independent of dictionary iteration order and of the transient
+///   `FilterState.modified` flag (which is not a query parameter).
+///
+/// Virtual / client-defined views (e.g. Stage 14's pinned "Downloaded" list) use
```
paulgessinger, line 25: Remove Stage reference

Persistence/Sources/Persistence/Models/QueryKey.swift
```diff
@@ -0,0 +1,55 @@
...
+    let ordering = (filter.sortOrder.reverse ? "-" : "") + filter.sortField.rawValue
+    // `.legacy` is used as a *canonical* encoding, not because the request will
+    // use it: the same search is spelled `title_content`/`title__icontains`
+    // pre-3.0 and `text`/`title_search` on 3.0+, and hashing whichever the
+    // backend happens to want would move every key when a server upgrades.
+    // Fixing the encoding keeps a query's identity stable across that upgrade.
+    let ruleLines =
+      FilterRule.queryItems(for: filter.rules(for: .legacy))
+      .map { "\($0.name)=\($0.value ?? "")" }
+      .sorted()
```
paulgessinger, line 41: We'll probably eventually retire this and force the migration, but I think we can defer that to whenever `.legacy` gets removed here.

Persistence/Sources/Persistence/Models/QueryOrderRow.swift
```diff
@@ -0,0 +1,59 @@
...
+    case orderStale = "order_stale"
+    case filledAt = "filled_at"
+  }
+}
+
+/// Public status of a cached query, surfaced to the list view-model: the
+/// server's `totalCount` (scrollbar extent), how many rows are locally present
+/// (`localCount`, reflects deletion gaps), and whether the cached order is stale
+/// under the active sort.
+public struct QueryStatus: Equatable, Sendable {
```
paulgessinger, line 50: The public struct is a pure subset of the other structs?

Persistence/Sources/Persistence/Database+DocumentDetail.swift
```diff
@@ -0,0 +1,58 @@
...
+/// Both follow the same network-first → write-through → offline-fallback shape
+/// as `document(id:)`: the caching repository forwards to the server, replaces
+/// the cached row on success, and serves the cached row on failure. Notes are
+/// document-keyed and mutable; file-metadata is version-keyed and immutable.
+extension Database {
+  // MARK: - Notes (mutable, document-keyed)
+
+  /// Replace a document's cached notes list (a row replace, not a per-note
+  /// merge — the endpoint and mutations always return the full list).
+  public func setNotes(_ notes: [DocumentNote], serverID: UUID, documentID: UInt) throws {
```
paulgessinger: Typed throws?

Persistence/Sources/Persistence/Database+Documents.swift
```diff
@@ -0,0 +1,217 @@
...
+        ).insert(db)
+      }
+      try setQueryMeta(
+        db, serverID: serverID, queryKey: queryKey,
+        totalCount: totalCount, orderStale: false)
+    }
+  }
+
+  /// Mark every cached query containing `remoteID` order-stale under its active
+  /// sort. v1 over-marks (any query the doc is a member of); the ordering
```
paulgessinger, line 77: What's v1?

Persistence/Sources/Persistence/Database+Documents.swift
```diff
@@ -0,0 +1,217 @@
...
+  /// corrects on the next fill / delta.
+  public func markQueriesOrderStale(containing remoteID: UInt, serverID: UUID) throws {
+    try writer.write { db in
+      try db.execute(
+        sql: """
+          UPDATE query_meta SET order_stale = 1
+          WHERE server_id = ? AND query_key IN (
+            SELECT DISTINCT query_key FROM query_order
+            WHERE server_id = ? AND remote_id = ?)
+          """,
```
paulgessinger: Can this be expressed in GRDB's Swift API?

AppShared/Sources/AppShared/Document/DocumentStore.swift

Yes, but it still has one caller: `set(repository:reload:)` calls it on a repository swap, and it still resets `tasks` and `lastSyncError` — both per-server. What it no longer does is drop documents (the list observes the DB now), which is what made the name and the `public` look wrong.

Done in `0ceb4a05` — made it `private` (nothing outside `DocumentStore` called it) and moved the explanation onto the function, so what it does and why is stated once instead of implied by an absence.

AppShared/Sources/AppShared/Document/DocumentStore.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `f775133a`, at the **top of #627** rather than here.

All 29 references are gone, verified repo-wide. That includes one in `DetailFillTests` that a source-only sweep would have missed — the ledger's count is what caught it. Three were reworded rather than deleted, since they pointed at stages that do not exist in the tree: `QueryKey` → "a future pinned Downloaded list", `Database+Connections` → names the APIs that now exist instead of promising them, and the `CachingRepository` plan-doc pointer keeps the citation but by section title only.

Why at the tip: these are doc comments on lines #617/#618/#627 rewrite, so a per-PR split would have conflicted in every direction for no benefit, and the stack merges as a unit. Closes item 10 in #656 (#662). Worth knowing it is **not** on `main` until #627 lands.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Right on both counts, and yes it is the standard shape — `UserDefaults` has no key-enumeration API, so sweeping `dictionaryRepresentation().keys` by prefix is the only way to clear a family of per-server keys.

No change made, because this whole enum stops existing two PRs up: #617 moves the watermark into `server_sync_state` in the database, where a per-server row is deleted by the `server` FK cascade and the prefix sweep is unnecessary. Not worth tidying code that is deleted upstack.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Already updated — on #617, where the change actually landed. The ascending-walk rewrite (`fix: make the changed-metadata delta resumable instead of stranding a backlog`) took this doc comment to "page forward from the per-server watermark…" in the same commit.

At *this* PR's head the `ordering=-modified` description is still accurate for the code below it, so there is nothing stale to fix here.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Not through the caching layer, no — which I think is what you are asking. Both real call sites reach past it: `reconcileDocumentDeletions` here and `fillLibrary` on #618 both call `wrapped.documentIDs(...)`, which goes straight to the inner repository and skips this override entirely.

Left in place deliberately. `documentIDs` has a default implementation on `Repository` that pages the *full* Tier-1 list and maps ids; delete this passthrough and any holder of an `any Repository` calling `documentIDs` silently drops onto that instead of `ApiRepository`'s `fields=id` projection. So it is unreachable today but load-bearing the moment anything calls it through the store's repository. The alternative — pointing the reconcile at `self.documentIDs` so it is genuinely exercised — is a behaviour-neutral change I did not want to make silently; say the word if you prefer it.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

This call site no longer cares: `ApiRepository.documentIDs` overrides the sort outright (`ordering=id` ascending) since the previous round's fix, precisely because ASN is NULL for most documents and paging a tie group that big drops ids across page boundaries. `3b5ba3f0` now does the same in the `Repository` **default** implementation, which had the identical bug and only mocks/previews reach.

Changing the default itself is a bigger question than it looks, so I recorded it rather than doing it — filed as **#669** (ledger item 27). `QueryKey` hashes the ordering, so moving `FilterState.empty`'s sort re-keys *every* cached query at once, and there is no cleanup path for keys that stop being reachable: the old rows sit in `query_order` / `query_meta` forever, and because they keep their documents "referenced" they also block `pruneUnreferencedDocuments`. Safe and cheap once that sweep exists; permanently stranding without it.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Skipping this one on purpose: #617 moves the watermark into `server_sync_state` and deletes all three of these `UserDefaults(suiteName:) ?? .standard` sites, so centralizing them here would be tidying code with two PRs left to live.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `0ceb4a05` — lifted into `fileMetadataVersionID(documentId:)`, so the call site reads as one named concept and the explanation sits on the helper.

Both optionals are load-bearing, which is why they collapse to a name rather than disappearing: the document row may not be cached yet, and the read itself can fail. Both fall back to the document id, which equals the root version id server-side.

AppShared/Sources/AppShared/Networking/QueryFillHandle.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Correct, and the comment is the part that is wrong — the value is used. `DocumentListViewModel:191` takes it off the handle and `DocumentList.swift:305` feeds it to `DocumentCountPill`, plus the `(totalCount ?? 0) > 0` empty-state gate. So it drives the **count pill**, never a scroll extent: the list renders only `documents`, the locally-loaded prefix, so the scrollbar follows what is loaded.

Done in `0a6ab7b3` — fixed here and at the seven other places repeating the same "scrollbar extent" claim (`QueryStatus`, `observeQueryStatus`, `queryStatus`, the `V4` header, `DocumentListViewModel`, `DocumentStore`, and the page-1 comment in `fillQuery`).

Networking/Sources/Networking/Api/ApiRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `0a6ab7b3` — six lines down to three. Kept the causal chain (non-unique ordering → ids dropped across page boundaries → reconcile reads them as deletions → evicts from cache), since that is the part that is not obvious from the code; dropped the worked ASN example.

Networking/Sources/Networking/Api/ApiRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Correct — the default only has the `DocumentSource` abstraction, and `fields=id` plus `pageSize: 25000` are URL-level, so it cannot express the cheap projection.

But one half of the override *is* expressible there, and should have been: the default pages `documents(filter:)` under the caller's ordering, so it carries the exact non-unique-paging bug the override was fixed for last round. Only mocks and previews reach it today, so nothing user-visible, but it was the same defect sitting in the fallback. Fixed in `3b5ba3f0` — the ordering is now stabilized in both places, and the doc comment names the projection as the part that genuinely cannot be hoisted.

Persistence/Sources/Persistence/Migrations/V4_CreateDocumentCache.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

We store JSON as TEXT and never query into it — deliberately. Everything that gets filtered or joined on is promoted to a real column (`title`, `asn`, `projection_level`, `detail_fetched_at`), precisely so SQLite never has to parse the blob and the ASN lookup can use its index. Same shape as the V3 element cache. If we ever do need to reach inside, the JSON1 functions work on TEXT as-is; JSONB would need SQLite 3.45+, which the iOS 17 floor predates.

The one thing we could add and do not have is a `CHECK (json_valid(data))`. Left out for now — the writer is our own encoder, so it would catch corruption rather than a bug — but happy to add it if you want the belt.

Persistence/Sources/Persistence/Migrations/V4_CreateDocumentCache.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

You are right, and it is worse than the comment suggests: at *this* PR's head nothing reads or writes `server_sync_state` at all. I checked every file at the head — the only mention is the migration itself. The watermark lives in app-group `UserDefaults` here, and both columns (`delta_watermark`, `library_coverage_at`) belong to #617.

Done in `3d74a665` (this PR) plus `270489a9` (#617): the table creation moves to #617's `V6`, where the first reader lives, and `V6` picks up the `tables` entry so the cache wipe still clears it. `V4` no longer creates a table its own build never touches.

One wrinkle worth stating plainly: within #617 the cursors commit sits *below* `V6`, so at that one intermediate commit the code references a table a later commit creates. The PR as a whole is coherent and migrations all run at launch, so the merged result is correct — fixing the intra-PR ordering properly would mean a new migration numbered `V6` and renumbering `V6→V7→V8→V9` across three PRs, which is a lot of churn for commit-level tidiness. Say so if you want it done properly.

Persistence/Sources/Persistence/Migrations/V5_CreateDocumentDetailCache.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Keeping them separate, per your call. For the record it would have been mechanically free — neither has shipped, and migration identifiers are strings, so dropping `v5_…` leaves a harmless gap with `V6`–`V8` keeping their names — but the only gain is a shorter migration list, against a rebase through four PRs of test references. Each migration arriving with the commit that needs it also reads better in review.

Persistence/Sources/Persistence/Models/QueryKey.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `f775133a` (#627), as part of the full sweep. Reworded rather than deleted, since "Stage 14" pointed at something that does not exist in the tree: it now reads "a future pinned Downloaded list", which is the actual meaning.

Persistence/Sources/Persistence/Models/QueryKey.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Agreed, and deferred. One thing to carry with that decision: retiring the fixed `.legacy` encoding changes the hash input, so **every** cached query key moves at once — and there is no cleanup for keys that stop being reachable, so the entire existing `query_order` / `query_meta` would be stranded permanently, taking `pruneUnreferencedDocuments` with it (orphaned membership rows keep their documents "referenced").

Filed as **#669** with your default-sort question, since both are gated on the same missing sweep. Once that exists, this is a safe change: the orphans get collected and the queries re-fill on next open.

Persistence/Sources/Persistence/Models/QueryOrderRow.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Not quite a subset — it is a projection across both tables plus a computed field. `totalCount` and `orderStale` come from `QueryMetaRow`, but `localCount` is a `COUNT(*)` over `query_order` rather than a stored column (which is what makes deletion gaps visible), and `filledAt` is deliberately not exposed. So it cannot be `QueryMetaRow` with fewer fields.

The doc comment said none of that; `0a6ab7b3` now spells it out, along with correcting the "scrollbar extent" claim on `totalCount`.

Persistence/Sources/Persistence/Database+DocumentDetail.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Yes — and it was a real invariant violation, not just a style gap. `Database+Elements.swift` states that every access goes through `Database/wrapping` so a GRDB failure reaches callers as a `DatabaseError`; the whole document-cache surface was untyped `throws` calling `writer.read`/`writer.write` directly, so raw `GRDB.DatabaseError`s escaped the package.

Done in `e3c1d06a`: every accessor in `Database+Documents`, `Database+DocumentDetail` and `Database+Maintenance` now declares `throws(DatabaseError)` and goes through `wrapping`. You were right that no caller-side updates were needed — a typed throw is accepted anywhere an untyped one was, and it builds clean at this PR's head and at the stack tip.

The comment on `wrapping` claiming a direct `writer.read` "fails to compile" is what let this drift, and it is now corrected: that only holds *inside an accessor that already declares the typed throw*. The compiler enforces nothing on one that stays untyped.

Still open, out of scope here: `Database+Connections.swift`'s four accessors have the same gap, but they are on `main` and belong to no PR in this stack. Recorded in #656.

Persistence/Sources/Persistence/Database+Documents.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

The plan document's name for the shipped variant of the source-of-truth design — meaningless from inside the code, which is the point of your comment. Done in `f775133a` (#627): it now says what it actually means, that the marking deliberately over-marks (every query the document is a member of, not only those whose sort field changed) because the ordering corrects itself on the next fill or delta either way.

Persistence/Sources/Persistence/Database+Documents.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Yes. Done in `fde53a21` — a `QueryOrderRow` subquery selecting `query_key`, then `QueryMetaRow.filter(...).updateAll(db, Column("order_stale").set(to: true))`. Same statement shape, and `DISTINCT` turns out to be unnecessary inside an `IN`.

Covered by the existing "markQueriesOrderStale flips the flag only for queries containing the doc" test, which passes on the rewrite.

PR #613, opened 2026-07-08
Implemented shortcuts functionality to automate common workflows

### LINKED ISSUE #610: Add shortcuts for the app
### Description

I wish this app exposed "shortcuts" in iphone (Is that the right term)? 

For example, I have a very common usecase/workflow where I take a photo of a receipt, open Swift Paperless -> import from photos -> change document type to "receipt" and add. 

I wish this action could be automated in shortcuts by importing the last photo in the camera roll as a receipt (Or any other document type). Ideally the app should expose shortcuts that allow fine grained control to suit complex workflows.

I can make this feature as a PR for @paulgessinger to review, if he would like my contribution.

DrakenKor:
# Add Shortcuts (App Intents) support

Closes #610

Adds an **Upload Document** action to the Shortcuts app, so document uploads can be automated — e.g. "take the latest photo from the camera roll and upload it as a Receipt" without opening the app. Can also choose the backend server to upload to.

## What's included

- `UploadDocumentIntent`: accepts a file (image or PDF) plus optional **Title**, **Document Type**, **Correspondent**, and **Tags**. Runs in the background and reports success or a descriptive error.
- Document type / correspondent / tag parameters are dynamic entities — their options are fetched live from the active server connection, with search support.
- An `AppShortcut` ("Upload document to Paperless") so the action is discoverable in Shortcuts and Spotlight.
- Small fix in `ApiRepository.create`: the multipart `title` field is now omitted when empty, letting the server derive the title from the filename.
- `AppShared/.gitignore`: same SwiftPM ignore rules as the sibling packages (was missing).

## Design notes

- Intents live in the main app target under `swift-paperless/Intents/` — no new target, no `project.yml` changes.
- Connection handling and uploads reuse the existing `ConnectionManager` / `ApiRepository` paths (same as the ShareExtension). Only the currently active connection is used.
- Static intent metadata (titles, parameter names) uses string literals because `appintentsmetadataprocessor` cannot evaluate the xcstrings-tool accessors; runtime strings (errors, success dialog) stay in `App.xcstrings`.
- Entity types are standalone, so future intents (search, edit, more upload parameters) can be added without refactoring.

## Screenshots

IMG_5351
IMG_5352

I tested it working locally with a test build on my iPhone 13 Pro.

## Testing

- Verified end-to-end on a physical iPhone (iOS 26): latest-photo-as-receipt automation uploads correctly, dynamic dropdowns populate from the server, file access works with security-scoped URLs from Shortcuts.
- Simulator build passes including AppIntents metadata extraction; all 112 package tests pass; `swift format lint --strict` clean.


paulgessinger:
hi @DrakenKor, I appreciate the effort. I was also thinking about this, and in your proposal the shortcut only ever uploads to the selected active server, right? My plan was to make the shortcut be able to select the target server explicitly. This will require some additional work. Maybe something *simpler* like this here is a good transitional solution, though.

DrakenKor:
@paulgessinger I'll add a shortcut step to select the server.

DrakenKor:
@paulgessinger I added server selection support (see screenshots) when creating the app shortcut.

DrakenKor:
@paulgessinger bump

paulgessinger:
@DrakenKor I'll get to this, but I think I want to land the rest of the offline PR stack first, so I can make sure it does not interfere.

DrakenKor:
@paulgessinger Sure thing. Just ping me after that's done and I'll rebase/resolve merge conflicts if any.

paulgessinger:
@DrakenKor I think we can proceed with this now. It'll need a rebase and a few adjustments I think:

- `ConnectionManager` now accepts a database, so the construction will need to change. This should ideally follow what happens in `ShareView.init` and `refreshConnection`.
- Element loading should go through `store.repository.tags()` etc, as that will ensure the results up-to-date from the db.
  -  suggestedEntities / entities(matching:): try? await store.sync() with a short timeout, then read. If the sync fails, return what's cached.  
  - entities(for:): read without syncing. If an ID is missing, sync once and read again. If it's still missing, throw an error saying the tag/type/correspondent no longer
- in `perform()` you could run an opportunistic `try await store.sync(userInitiated: true)` while the document is uploading.

Let me know what you think!

DrakenKor:
@paulgessinger I've updated my changes according to your recommendations. Just doing a final rebase because your main had another push while I was making changes.

PR #614, opened 2026-07-11
ci: Harden and speed up GitHub Actions workflows

paulgessinger:
Hardens and cleans up the CI setup. No behavioral change to what gets built/deployed — this is supply-chain, security, and cost hygiene.

## Changes

**Pin all actions to commit SHAs + enforce it** (`3d6695327`)
- Every `uses:` across all workflows now pins to a full commit SHA with a version comment (done via [`pinact`](https://github.com/suzuki-shunsuke/pinact)). This also normalized previously inconsistent versions (checkout was a mix of v3/v4/v5, deploy-pages v2/v4, etc.).
- Added the [`zizmor`](https://github.com/zizmorcore/zizmor) pre-commit hook so the existing `prek` lint job fails on any future unpinned action.
- Fixed zizmor's safe findings: workflow-level `contents: read` on build, scoped `pages`/`id-token: write` to the docs `deploy` job only, and `persist-credentials: false` on checkouts that don't push. `beta.yml` is left intact (fastlane pushes the build-number bump via the checkout token) with its two findings suppressed inline + rationale.

**CI hygiene** (`edf3847a3`)
- `docs.yml` now builds on `main` + PRs instead of every branch push.
- Added `timeout-minutes` to every job so a hung runner can't burn the 6-hour default.

**SwiftPM caching** (`d2740d70a`)
- Test matrix caches each package's `.build` (keyed on its `Package.swift`).
- Build job resolves into a fixed `SourcePackages/` dir and caches it (keyed on all `Package.swift` + `project.yml`). Keys include runner OS + Xcode version.

**Remove outdated static Pages workflow** (`987df705e`)
- `static.yml` deployed the `pages` branch to the same `github-pages` environment that `docs.yml` now serves from `main` — redundant.

## Note
Caching effectiveness (hit-rate / time saved) needs a real CI run to confirm; the keys and paths are sound but unmeasured locally.

https://claude.ai/code/session_015DccxfKNt2ufN5Ei3fDz3D


paulgessinger:
@codex

PR #617, opened 2026-07-11
feat: Entire library is offline

paulgessinger:
This pull request introduces a comprehensive set of changes to support and implement a new "Offline Browsing Mode" for document caching, as well as improvements to proactive offline fill and cache management. The main themes are: adding per-server offline browsing modes, implementing proactive full-library and detail fills, enhancing cache observation and UI feedback, and improving cache-clearing logic. Additionally, there are new localization helpers for UI presentation.

**Offline Browsing Mode & Proactive Fill:**

* Added the `OfflineBrowsingMode` enum and logic to persist and retrieve per-server offline browsing preferences, allowing users to choose between "recently browsed" and "entire library" caching strategies. The default mode is adaptive to library size. (`AppShared/Sources/AppShared/Model/OfflineBrowsingMode.swift` [[1]](diffhunk://#diff-5d1afb6de44daaf3b92dcca60ba7edcda9519f130801fe5dc1cd88577b85014cR1-R34) `AppShared/Sources/AppShared/Networking/CachingRepository.swift` [[2]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR128-R132) [[3]](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feR157-R163)
* Implemented proactive full-library fill (`fillLibraryIfEnabled`) and per-document detail fill (`fillDocumentDetailsIfEnabled`) in `DocumentStore`, gated by the offline browsing mode and network conditions. These functions ensure the entire library and its details are available offline when enabled. (`AppShared/Sources/AppShared/Document/DocumentStore.swift` [AppShared/Sources/AppShared/Document/DocumentStore.swiftL724-R848](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L724-R848))

**Live Observation & UI Feedback:**

* Added live-observed properties to `DocumentStore` for library coverage timestamp, per-view sync errors, and cached document count, with observation tasks that update these properties as the underlying data changes. This enables the UI to provide accurate, live feedback about offline status and cache health. (`AppShared/Sources/AppShared/Document/DocumentStore.swift` [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R82-R103) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R138-R150) [[3]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70R169-R214)

**Cache Management Improvements:**

* Improved cache clearing to also reset the library coverage marker and clarified documentation around what gets wiped and when the next fill or reconcile is triggered. (`AppShared/Sources/AppShared/Document/DocumentStore.swift` [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L744-R866) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L760-L762)
* Refactored the previous `DocumentDeltaWatermark` logic into a new `LibraryCoverage` freshness policy, which determines when a proactive fill should re-run based on a configurable age threshold. (`AppShared/Sources/AppShared/Networking/CachingRepository.swift` [AppShared/Sources/AppShared/Networking/CachingRepository.swiftL32-R46](diffhunk://#diff-8d55fb15fbb2a18b28da6f85fd2ca2f25674f8a8af4c8950bce9367743ec85feL32-R46))

**Networking & Sync Enhancements:**

* Updated sync and reconcile operations to categorize network transfers (for better UI/metrics) and to ensure saved view memberships are rebuilt after changes, so offline lists remain accurate. (`AppShared/Sources/AppShared/Document/DocumentStore.swift` [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L406-R479) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L724-R848)
* Clarified and improved deletion logic to explicitly prune all references to deleted documents, ensuring cache consistency. (`AppShared/Sources/AppShared/Document/DocumentStore.swift` [AppShared/Sources/AppShared/Document/DocumentStore.swiftL286-R358](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L286-R358))

**Localization:**

* Added localization helpers for both `OfflineBrowsingMode` and `TransferCategory` to provide user-friendly, localized names in the UI. (`AppShared/Sources/AppShared/Model/Localization/OfflineBrowsingMode+localizedName.swift` [[1]](diffhunk://#diff-6fbb0876b0139a4076d207e2756b4a22353264b4d0f2d57e9899b26986053b14R1-R17) `AppShared/Sources/AppShared/Model/Localization/TransferCategory+localizedName.swift` [[2]](diffhunk://#diff-993efac34529310ba56c5260c14b84f1d021e7a77d29944b6fba91995d8bc277R1-R21)

These changes collectively enable a much more robust, flexible, and user-transparent offline experience, especially for users with large or multiple libraries.

## Review round (2026-08-16)

Three follow-up commits at the end of this PR's range, from the review of #600
([findings](https://github.com/paulgessinger/swift-paperless/pull/600#issuecomment-5306727572); context in [this comment](https://github.com/paulgessinger/swift-paperless/pull/617#issuecomment-5307576413)).

| Commit | What it fixes |
|---|---|
| `carry the query_order unique key through the V6 rebuild` | `V6` recreates `query_order` to drop the `document` FK, and a rebuilt table doesn't inherit `V4`'s new uniqueness constraint — so the no-duplicates fix on #600 silently evaporated the moment this PR landed. Verified by the test failing at the stack tip with the `V4`-only version. Also switches `replaceQueryOrder` to `.insert`. |
| `make the changed-metadata delta resumable instead of stranding a backlog` | R3δ walked newest-first and advanced the watermark to the newest change even when it stopped early on `deltaCap`, so a bulk edit larger than the cap left most rows permanently stale. Now walks ascending from the watermark, committing the cursor per page. |
| `carry the query_order position conflict clause through the V6 rebuild` | Same rebuild problem for the `.replace` clause added on #600. |

Two things worth a reviewer's attention, both called out in the linked comment: the delta's
**known bound** (more than `deltaCap` documents sharing one `modified` timestamp stalls the
cursor — needs 1000+ rows inside a millisecond), and a **decision still open** on whether the
delta or `reconcileSavedViewMembership` owns ordering repair under *Recently browsed* (#656
item 20).

One conflict arose and was resolved in the commit that owned it: #618's `Logger.shared` →
`Logger.sync` rename lands on the log line the delta rewrite relocates into the page loop.

### Review round 2 (2026-08-16)

One further commit, from Paul's comment round on #600: **`create server_sync_state with the cursors that first use it`**. `V4` was creating that table for a build with no cursors at all — both columns are read and written here — so its creation moved into this PR's `V6`, which also picks up the `tables` entry so the cache wipe still clears it.

One wrinkle, stated plainly: within this PR the cursors commit sits *below* `V6`, so at that one intermediate commit the code references a table a later commit creates. The PR as a whole is coherent (migrations all run at launch), but the intra-PR ordering isn't. Fixing it properly means a new migration numbered `V6` and renumbering `V6→V7→V8→V9` across three PRs.

## Manual test checklist

Nothing in this PR has been driven in the simulator — everything is build- and
unit-test-verified only (Common 61, DataModel 230, Networking 217, Persistence 134).
The items below are the ones automated tests can't reach. Grouped roughly by setup cost;
the first two sections are the ones that would hurt most if wrong.

### Upgrade from a shipped build *(highest risk — one-shot, not repeatable without a clean install)*

`V6` rebuilds `query_order` (create / copy / drop / rename), `V8` adds a `server`
column, and `V9` adds two `document` columns and backfills them from the JSON blob.
All three only ever run against real data once, on the upgrade.

- [x] Install the current App Store / TestFlight build, log in, browse a few saved views and the document list so `query_order`, `query_meta` and the document cache are populated.
- [x] Upgrade to this branch **without deleting the app**. The document list and every saved view still show the same documents, in the same order, at the same count.
- [x] Saved views still open offline (airplane mode) — i.e. the `query_order` copy carried the rows across, not just the schema.
- [x] Existing servers are on *Recently browsed* and nothing started downloading by itself.
- [x] Documents that had notes before the upgrade still get them fetched (this exercises `V9`'s `notes_count` backfill — if it silently produced zeroes, the detail fill would simply never ask for any notes again).
- [x] A server configured before the v2 → GRDB connection import still works (this exercises `V2_ImportLegacyConnections`, which this PR had to pin to V1's column list — a regression here shows up as a failed migration at launch, i.e. the app not starting).

### Document list ordering *(the regression this round fixed)*

The membership sweep used to rewrite every cached list to document-id order on every
reconcile. It runs on a 5-minute throttle and on pull-to-refresh, so the window is short.

- [x] Set the document list to a non-id sort (e.g. *Created*, descending). Pull to refresh. Order is unchanged.
- [x] Wait ~5 minutes with the app foregrounded, or pull to refresh a few times. Order is still unchanged — not silently re-sorted oldest-id-first.
- [x] Same for a saved view with its own sort.
- [x] With *Entire library* on, let a full fill complete, then pull to refresh. Order survives the membership sweep.
- [x] A saved view sorted by a **non-unique** field with many ties (e.g. *Added* on a bulk import) pages without dropping or repeating documents — scroll to the bottom and compare the count against the server.

### Skeleton rows

- [x] Enable *Entire library*, let the fill start, force-quit mid-fill, relaunch in airplane mode. Lists show a mix of loaded and redacted placeholder rows.
- [x] **In that mixed state**, skeleton rows are the same width and indentation as loaded rows (they were missing `listRowInsets` and rendered visibly narrower).
- [x] Skeleton rows are not tappable and issue no thumbnail requests.
- [x] Coming back online, skeletons fill in without the list jumping or duplicating rows.

### Offline browsing mode

- [x] A newly added server defaults to *Recently browsed*, regardless of library size, and does not start downloading.
- [x] On a server with fewer than 2,000 documents, the picker footer shows the "small enough to keep on your device in full" recommendation. (It reads the default list's cached total, so it may need the list opened once first.)
- [x] On a larger server, no recommendation appears.
- [x] Switching **to** *Entire library* starts a fill immediately (on Wi-Fi).
- [x] Switching **away** from *Entire library* asks for confirmation first, and the dialog says downloaded documents will be removed. Cancelling leaves the setting on *Entire library* and deletes nothing.
- [x] Confirming the downgrade shrinks *Cached documents*, and *Last full sync* resets to Never.
- [x] Start a fill on *Entire library*, then switch to another server mid-fill. The old server's fill stops — the cached count for the server you left should not keep climbing.
- [x] Start a fill, then downgrade to *Recently browsed* mid-fill. The fill stops rather than writing back rows the reclaim just removed; *Cached documents* settles low and stays there.
- [x] Toggle the mode back and forth quickly a few times. Only one fill runs — Activity shouldn't flicker between stages, and the progress bar shouldn't restart repeatedly.
- [x] Immediately after a downgrade, force-quit the app. On relaunch the cache is not half-reclaimed — saved views untracked but every document still on disk was the old failure mode.
- [x] Re-enabling *Entire library* after a downgrade actually re-fills (rather than reading a stale coverage marker and doing nothing).

### Progress and status accuracy

- [x] During a library fill, *Activity* shows "Downloading library…", the progress bar advances, and the saved view name appears beside it.
- [x] During the detail fill, *Activity* shows "Downloading document details…" with its own progress — it used to read "Idle" for the entire pass.
- [x] During a reconcile, *Activity* shows "Checking for changes…".
- [x] *Sync now* is disabled while a fill is running, and does not re-enable mid-fill when an overlapping pass finishes.
- [x] *Last refreshed* updates after a successful refresh **and repaints without leaving the screen**.
- [x] Point the app at an unreachable server (VPN off, LAN-only server) and refresh: *Last refreshed* does **not** advance. It only stamps on success.
- [x] *Cached documents* is formatted for the locale (thousands separator), not a bare integer.
- [x] VoiceOver reads each status row as one element (label + value), not two unrelated ones.

### Network gating

- [x] On cellular with *Entire library* on and the toggle off, *Activity* reads "Waiting for Wi-Fi" — not "Idle".
- [x] Turning on *Sync over cellular* lets the fill proceed on cellular.
- [x] The toggle is per-server: set it on one connection, switch to another, confirm it's independent.
- [x] With **Low Data Mode** on for the current network, the fill stays paused *even with* *Sync over cellular* enabled.
- [x] Interactive actions (opening a document, pull-to-refresh, search) still work normally on cellular regardless of the toggle.
- [x] On Wi-Fi but with the **server unreachable**, a fill attempt leaves *Last full sync* at its previous value — it must not stamp "just now" over a cache it didn't fill.

### Restricted (non-superuser) account

Needs a paperless-ngx user without note-view permission, and ideally one without
permission on some documents.

- [x] With *Entire library* on, the detail fill does not hammer `/notes/` — check the server log for a burst of 403s repeating on every foreground.
- [x] Documents the user can't change still show edit affordances correctly (this PR's `full_perms` switch changed where that answer comes from).
- [x] Opening a document whose permission was revoked server-side behaves sanely.

### Error surfacing

- [x] *Sync now* against an unreachable server shows an error. It used to fail completely silently.
- [x] A saved view the server rejects appears under *Sync problems*, with a relative timestamp of when it failed, and does not block the other views from syncing.
- [x] Opening a document while the server is unreachable does **not** remove it from the list.

### Data transfer meter

- [x] Figures are plausible for the traffic — they now count wire bytes, so a library fill should read *lower* than before (the API gzips well), not higher.
- [x] Browsing entirely from cache (airplane mode) adds nothing to the total.
- [x] *Reset statistics* zeroes the counters and resets the "since" date.
- [x] Backgrounding the app doesn't lose the last few seconds of counted traffic.

### Localization

- [x] Every new string on the Offline & Sync screen renders (no raw keys), including the progress counter and the confirmation dialog.
- [x] Nothing overflows or truncates badly at the largest Dynamic Type size.

**Stack** (bottom to top):

<details>

- ~~#617 feat/offline-entire-library~~ (merged)  👈

</details>

- #618 feat/multi-server-sync
- #627 feat/background-sync


paulgessinger:

## Two review fixes landed here that originate from #600's review

Both were found reviewing [#600](https://github.com/paulgessinger/swift-paperless/pull/600#issuecomment-5306727572) but belong on this PR.

### 1. `V6` was dropping a uniqueness constraint added in `V4`

`query_order` has no uniqueness on `remote_id`, so a document the server repeats across adjacent pages (an insert or delete shifting the page offsets mid-fill) is stored at two positions and the windowed replay emits it twice — a duplicate `ForEach` id in the list, plus an inflated `localCount`. #600 fixes that with `uniqueKey([server_id, query_key, remote_id], onConflict: .ignore)` in `V4` and a plain `.insert` in `writeQueryPage`.

`V6_DropProjectionAndQueryOrderFK` rebuilds `query_order` (create-new / copy / drop / rename, because SQLite can't alter an FK in place), and **a rebuilt table doesn't inherit the constraint**. With the fix in `V4` only, the new test passes on #600 and then fails at the stack tip returning `[1, 2, 3, 3, 4]`. So the constraint is declared in `V6`'s table too.

Note it's orthogonal to the FK this migration deliberately removes: the FK required `remote_id` to *reference a cached document*, the unique key requires it to *appear at most once per query*. Skeleton rows stay legal. With the FK gone this is in fact the only remaining constraint on `remote_id`.

`replaceQueryOrder` is switched to `.insert` for the same reason — it writes an ordered id list from the paged Tier-0 sweep, which can repeat an id the same way, and `upsert` would *move* the existing entry rather than skip the repeat.

### 2. R3δ stranded any backlog larger than `deltaCap`

The delta walked newest-first and advanced the watermark to the newest change seen — including when it stopped early on the cap. Everything below the cap was never revisited, so a server-side bulk edit larger than the cap left most affected rows stale until the user refreshed that list or opened the document.

No watermark value fixes the descending walk: a high-water mark only moves up, so it cannot express "applied `[oldest, newest]` but not `[watermark, oldest)`". (My first suggestion on #600 was exactly that, and it's wrong — corrected there.)

The catch-up now walks **oldest-first from the watermark**, committing the cursor after each page, so `deltaCap` becomes a per-pass budget: a pass that stops early — capped, cancelled, or killed on a background time budget — leaves the watermark at the last document applied and the next pass resumes there. That last case matters given the background sync in #627.

Two things to look at while reviewing:

- The server-side `modified` bound is date-granular and exclusive (`modified__date__gt`, no `gte`), which `FilterState` already compensates for by widening an inclusive start by a day. A pass therefore re-fetches from the start of the watermark's day.
- **Known bound:** if more than `deltaCap` documents share one `modified` timestamp, the cursor can't advance past them and successive passes re-apply the same group. That needs 1000+ rows inside a single millisecond (the decoder truncates to milliseconds), so I left it rather than adding a tie-breaker — but it's the one case where resumability doesn't hold, and it's worth a second opinion.

It landed here rather than on #600 because this PR's `reconcileDocumentChanges` is a materially different function — `entireLibrary` gating, `fullPerms`, `.full` projection, `invalidateNotes`. Rewriting it on #600 conflicted from this PR's first commit onward; here it folds into the finished version cleanly.

### Verification

App builds clean at the stack tip, Persistence 122/122, no conflicted commit in the stack. One conflict did arise and was resolved in the commit that owned it: #618's `Logger.shared` → `Logger.sync` rename lands on the log line the delta rewrite relocates into the page loop — resolved by keeping both, with #618's orphaned copy of the old trailing `setDeltaWatermark` block dropped. Nothing was driven in the simulator.

paulgessinger:

One item from [#600's review](https://github.com/paulgessinger/swift-paperless/pull/600#issuecomment-5306727572) belongs here, but it **needs a decision before any code** — ledger #656, item 20.

**R3δ refreshes rows without marking the query order stale.** `update(document:)` deliberately calls `markQueriesOrderStale(containing:)` because ordering isn't recomputed offline; `reconcileDocumentChanges` applies the same class of change and calls nothing. Rename a document server-side and it keeps its old position under an active sort.

The reason this isn't just a missing line: **this PR moved ordering repair out of the delta on purpose**, into `reconcileSavedViewMembership`, which rebuilds `query_order` wholesale from the Tier-0 id projection and runs after the R3δ pass. Adding the mark back into the delta re-introduces a parallel mechanism for something that sweep now owns. But the sweep self-gates on `offlineBrowsingMode == .entireLibrary`, so **`.recentlyBrowsed` is genuinely uncovered**.

So the question is which one owns ordering repair, not whether to add a call. Worth noting it's inert either way today — nothing reads `orderStale`; `startStatusObservation` consumes only `totalCount` — so there's no urgency, and it should be settled together with ledger item 23 (an append silently clears the same flag).

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -723,14 +798,56 @@ extension DocumentStore {
...
+        try await backend.reconcileSavedViewMembership()
+      }
     } catch {
       Logger.shared.info("Document reconcile failed (suppressed): \(error)")
     }
   }
 
+  /// When the document reconcile sweep (R2/R3δ/membership) last ran, for the
+  /// Offline & Sync status screen. `nil` until the first reconcile this session.
+  public var lastReconcileAt: Date? { lastDocumentReconcile }
```
paulgessinger: Why not up with the other members, and why not `private(set)`?

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -723,14 +798,56 @@ extension DocumentStore {
...
+
+  /// Proactive *Entire library* fill, gated by the setting and an unmetered link.
+  /// Foreground-only; soft-fail (offline-tolerant). `force` ignores the freshness
+  /// marker (e.g. the user just enabled the setting). A no-op when the setting is
+  /// *Recently browsed*, the link is metered, or there's no caching backend.
+  public func fillLibraryIfEnabled(unmetered: Bool, force: Bool = false) async {
+    guard unmetered, let backend = repository as? any CachingBackend,
```
paulgessinger, line 865: Whether the sync runs on wifi only or also on cellular should probably be a user setting?

Can we access the low-data-mode signal for the connection?

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -759,9 +878,6 @@ extension DocumentStore {
     }
     // Nuke memory + disk image cache.
     imagePipeline.cache.removeAll()
-    // Per-server changed-metadata delta watermarks (regenerable sync state):
-    // clear them so the reconcile re-baselines over the now-empty cache.
-    DocumentDeltaWatermark.clearAll()
```
paulgessinger, line 764: Why does this go away?

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -29,24 +29,21 @@ import Persistence
...
-    }
+/// The fill is skipped while the last-completed timestamp (in `server_sync_state`)
+/// is younger than ``maxAge``, so it runs once and then re-runs only as a periodic
+/// backstop — in particular a cold launch after a long quiet period (few/no
+/// activation sweeps) finds a stale marker and re-fills. Non-generic so the
+/// `static` constant is legal (it wouldn't be on the generic `CachingRepository`).
+enum LibraryCoverage {
+  /// Re-run the full fill at most this often as a backstop (the cheap activation
+  /// sweeps keep things current in between).
+  static let maxAge: TimeInterval = 7 * 24 * 60 * 60
```
paulgessinger: Let's discuss the schedule frequency again

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -179,42 +234,137 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
...
+    // failing view (recorded above) would otherwise pin "last full sync" at
+    // Never forever. Only cancellation — an interrupted run — skips the marker,
+    // and that bails out via the `throw` above before reaching here.
+    try? database.setLibraryCoverageAt(Date(), serverID: serverID)
+  }
+
+  // How many notes/metadata requests a single detail-fill pass makes before
+  // yielding, so a cold large library spans several foregrounds (mirroring the
+  // R3δ `deltaCap`) rather than one marathon of per-document round-trips.
+  private let detailFillCap = 500
```
paulgessinger: Maybe we don't actually want this? The first cold fill can continue, but it should be reported in the sync screen.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -179,42 +234,137 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
...
+
+  public func fillDocumentDetails() async throws {
+    guard offlineBrowsingMode == .entireLibrary else { return }
+
+    // Free step: every zero-note document gets an empty notes row from the list
+    // payload's count — no request — so it renders "no notes" offline and drops
+    // out of the fetch set below.
+    let seeded = (try? database.seedEmptyNotesForZeroCountDocuments(serverID: serverID)) ?? 0
```
paulgessinger, line 417: And this is because we write a single not entry for each document, containing all notes, rather than writing all notes as individual database rows. What does that buy us?

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -524,70 +688,142 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
     try database.deleteDocuments(serverID: serverID, removedIDs: Array(removed))
   }
 
-  // The number of changed documents one delta pass will apply before stopping
-  // (a runaway guard; the next pass continues from the advanced watermark).
+  // How many changed documents one delta pass applies before yielding. A budget,
+  // not a correctness boundary: the walk is resumable, so whatever is left over
+  // is picked up by the next pass from the advanced watermark.
   private let deltaCap = 1000
```
paulgessinger: Same here, maybe no cap with reporting is the better approach?

DataModel/Sources/DataModel/OfflineLibrarySize.swift
```diff
@@ -0,0 +1,29 @@
...
+//  OfflineLibrarySize.swift
+//  DataModel
+//
+
+import Foundation
+
+/// Pure decision logic behind the size-adaptive default for
+/// `AppShared.OfflineBrowsingMode`. Lives here (not in `AppShared`, which has
+/// no test target of its own) so the comparison gets a real unit test.
+public enum OfflineLibrarySize {
```
paulgessinger, line 11: Should this also have the fill/reconcile caps, if we decide to keep them?

Persistence/Sources/Persistence/Database+Documents.swift
```diff
@@ -93,20 +112,109 @@ extension Database {
...
+  /// Deletes every `document` row for `serverID` no longer referenced by any
+  /// `query_order` row for that server (the anti-join mirror of the
+  /// skeleton-read `queryWindowSQL`), plus its detail-cache siblings. Called
+  /// when a server's `OfflineBrowsingMode` transitions `.entireLibrary` →
+  /// `.recentlyBrowsed`, to reclaim documents that were only cached because of
+  /// the proactive fill. "Referenced" today means only `query_order`
+  /// membership — pinning (a second future exemption) doesn't exist yet.
+  /// Returns the number of documents removed (for logging/tests).
+  @discardableResult
+  public func pruneUnreferencedDocuments(serverID: UUID) throws -> Int {
```
paulgessinger: Wrap GRDB errors + typed throws for all database methods?

Persistence/Sources/Persistence/Database+Documents.swift
```diff
@@ -93,20 +112,109 @@ extension Database {
...
+    try writer.write { db in
+      let orphanIDs = try UInt.fetchAll(
+        db,
+        sql: """
+          SELECT d.id FROM document d
+          LEFT JOIN query_order q
+            ON q.server_id = d.server_id AND q.remote_id = d.id
+          WHERE d.server_id = ? AND q.remote_id IS NULL
+          """,
```
paulgessinger: Can this be written in GRDB's Swift API?

Persistence/Sources/Persistence/Database+Documents.swift
```diff
@@ -177,10 +292,29 @@ extension Database {
...
+  /// LEFT JOIN so a `query_order` id with no `document` row yields a skeleton
+  /// (NULL `document` columns); `q.remote_id` always carries the id.
   static let queryWindowSQL = """
-    SELECT d.* FROM query_order q
-    JOIN document d ON d.server_id = q.server_id AND d.id = q.remote_id
+    SELECT q.remote_id, d.* FROM query_order q
+    LEFT JOIN document d ON d.server_id = q.server_id AND d.id = q.remote_id
     WHERE q.server_id = ? AND q.query_key = ?
     ORDER BY q.position
     LIMIT ? OFFSET ?
```
paulgessinger, line 378: Likely not possible to express in GRDB's Swift API?

paulgessinger (commented):
- Make the comments more concise
- Is there any way we could surface progress from the fill and reconcile passes so we can show a stage label and a progress bar?
- How can I verify the skeleton presentation that was added here?

AppShared/Sources/AppShared/Document/DocumentStore.swift

Done in `bc19400` — moved up with the other observed status members and made `public private(set) var lastReconcileAt: Date?`.

Both halves of the question turned out to matter. It wasn't just placement: the backing store was `@ObservationIgnored`, so writing it invalidated nothing and the "Last refreshed" row only repainted when some *other* observed property happened to fire. On a quiet server it read "Never" indefinitely.

While in there, two related fixes: it was stamped *before* the sweeps ran, inside a suppressing catch, so a server failing every reconcile still showed a fresh timestamp — it now advances only on success. The throttle clock is a separate private field, since that one does have to advance on every attempt.

AppShared/Sources/AppShared/Document/DocumentStore.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `f86a688` — added a per-server **Sync over cellular** toggle on the Offline & Sync screen, stored on the connection record beside `offlineBrowsingMode` (new `V8` migration), defaulting off.

On the low-data-mode signal: we already read it. `NWPath.isConstrained` **is** Low Data Mode, and `isExpensive` is cellular/hotspot; the gate was `!isExpensive && !isConstrained`. The policy now lives in one place, `NetworkMonitor.allowsProactiveSync(syncOverCellular:)`, and Low Data Mode always wins — it's an explicit system-wide instruction from the user, so a per-server opt-in doesn't override it.

Worth knowing: neither signal means *reachable*. Being on Wi-Fi with the server unreachable looked identical to a healthy link, which is how a fully failed fill was stamping the coverage marker (fixed in `228ca77`). And enabling *Entire library* on cellular used to do nothing at all, silently — there's now a "Waiting for Wi-Fi" activity state instead of "Idle".

AppShared/Sources/AppShared/Document/DocumentStore.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Nothing goes away — it moved, and this call became redundant.

The delta watermarks used to live in app-group `UserDefaults`, which is why `wipeLocalCache` had to clear them explicitly. This PR's "store per-server sync cursors in the database" commit moves them into `server_sync_state`, and `Database+Maintenance.clearCache()` builds its table list from `V6_DropProjectionAndQueryOrderFK.tables`, which includes `server_sync_state`. So the DB wipe already clears both cursors.

Verified both ends at this PR's head. The doc comment on `wipeLocalCache` spells the new route out, so the behaviour is unchanged — the cache wipe still re-baselines the reconcile and re-runs the fill.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `228ca77` — `LibraryCoverage.maxAge` is now 24 hours rather than 7 days.

Reasoning for daily: the delta and the membership sweep carry freshness between runs, so a full pass mostly re-confirms membership and is cheap. What the interval really governs is how long any gap *they* miss can persist — and a week was a long time to be wrong.

It also bounds a failure mode found in this round: a pass where every view failed was still stamping the marker, so a fill that cached nothing suppressed all retries for the whole window while the screen reported a fresh full sync over an empty cache. That's fixed independently (the marker only advances if at least one view succeeded), but the shorter window limits the blast radius of anything similar.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Agreed, and done — `712a0973` removes `detailFillCap`, `bc19400` adds the reporting.

The pass now runs to completion; only cancellation (backgrounding, a server switch) stops it early, and because it's driven off what's still missing it resumes rather than restarts. The Offline & Sync screen gained a stage label and a determinate progress bar, so a first cold fill is visible instead of hidden behind a budget.

Removing the cap also fixed two things the cap was causing rather than containing:

- **Head-of-line blocking.** `prefix(500)` over an unordered query meant a permanently-failing document sat at the head forever. A non-superuser without note permission burned 500 doomed `/notes/` requests every single foreground and never cached a single note. The notes leg is now gated on the permission, and failed ids are skipped for the session.
- The queries themselves were `fetchAll` over the whole `document` table with a JSON decode per row — three of them per foreground, one inside the write transaction holding the writer lock. They're SQL predicates now (`json_extract`, anti-joins, `ORDER BY id`), so the work is proportional to what's actually missing.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Nothing, beyond matching the endpoint — and I'd keep it.

The `/notes/` endpoint returns the whole list, and create/delete also return the updated whole list, so there's no partial-update path a per-note table would serve. Every write would still be delete-all-plus-insert-N, and reads would gain a join. `DocumentNoteRecord`'s own doc comment already makes this argument.

What the blob shape genuinely costs is *predicates*: "which documents still need notes fetched" can't be a plain SQL filter over a JSON array. That's why `documentIDsNeedingNotesFetch` used to decode every document row in the table. But that's fixable inside the current shape, and is fixed in `712a0973` — `json_extract(d.data, '$.notesCount') > 0` plus a `NOT EXISTS` anti-join against `document_note`, with no decode at all.

So the one real drawback has a cheap answer, and per-note rows would buy nothing back.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `64ace961` — `deltaCap` is gone, with progress reported instead.

You were right that the cap was the wrong tool, and it was worse than just conservative: it counted *applied* documents. Under *Recently browsed* nearly everything is filtered out before it counts, so it barely bounded anything. Under *Entire library* it could wedge the cursor permanently — a run of documents sharing one `modified` value exhausts the budget without the cursor advancing, so the watermark never moves and nothing newer is ever reached again.

That's not the exotic case it sounds like: a single server-side bulk `UPDATE` stamps every affected row with an identical timestamp (paperless-ngx's `reflect_doclinks` does exactly `.update(modified=timezone.now())`), so ties are systematic rather than a millisecond coincidence. Running to exhaustion always advances the cursor to the true newest.

Separately, the walk was offset-paging a **non-unique** ordering — the same hazard `documentIDs` documents and works around by forcing `id` order. Rows could fall between page boundaries and, once below the committed watermark, never be revisited. `ApiRepository.documents(filter:)` now appends `id` as a secondary ordering key, which fixes it for `fillQuery` too.

DataModel/Sources/DataModel/OfflineLibrarySize.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Moot now, in the good way — both caps are gone (`64ace961`, `712a0973`), so there's nothing left to consolidate.

`OfflineLibrarySize` did end up with a changed role though. Since the size-adaptive default is out (a new server now starts at *Recently browsed* whatever its size — a whole-library download shouldn't happen before the user has seen the setting exists), `isSmall` had no production caller left. Rather than delete it, `0809d27` uses it for a hint under the picker: shown only while the server is on *Recently browsed* and the library is under the threshold. The count comes from the default list's cached `query_meta`, so it costs no request.

`recentlyBrowsedDefaultListCap` stays where it is — it's the one number that genuinely still answers "how light should *Recently browsed* stay", and it's used by the downgrade reclaim.

Persistence/Sources/Persistence/Database+Documents.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done in `3bb9ec8` — all 22 remaining accessors converted.

You'd flagged the real shape of it: #600 established the convention and applied it to the document-cache accessors that existed then, and everything added since quietly went back to a bare `throws` with a direct `writer.read`/`writer.write`. That's the whole of `Database+SyncState`, `Database+QuerySyncError`, the `query_order` writers, the detail-fill predicates — plus the four in `Database+Connections` that predate the stack (the ledger had those recorded as knowingly out of scope; they're included here since the sweep was happening anyway).

The practical effect was that those errors missed `DatabaseError+DisplayableError`, so anything reaching `errorController` surfaced as raw GRDB text. No behaviour change otherwise — a typed throw is accepted wherever an untyped one was, and `wrapping` rethrows an existing `DatabaseError` unchanged.

Persistence/Sources/Persistence/Database+Documents.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

It can, but it needs plumbing that doesn't exist yet, so I left it as SQL.

The anti-join is expressible through the query interface, but only via a declared `HasMany`/`BelongsTo` association between `DocumentRecord` and `QueryOrderRow` — GRDB won't join two records that aren't related. That's real association machinery added for one query, and `joining(optional:)` + `filter(...IS NULL)` reads no more clearly than the four-line `LEFT JOIN` does.

The balance also shifted in the other direction this round. The detail-fill predicates moved *to* raw SQL in `712a0973`, because they need `json_extract` on the document blob and `json_each` over the versions array to reproduce `currentVersionID` — neither of which the query interface can express at all. So the file now has a consistent rule worth stating: query interface for row-shaped work, SQL where the predicate reaches into the JSON payload or joins across unrelated records.

Happy to add the associations if you'd rather have uniformity — it's contained, just not obviously a win.

Persistence/Sources/Persistence/Database+Documents.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Right, not usefully. This one is a genuine outlier.

GRDB *can* do a left join with `including(optional:)`, but this query's whole point is the shape of the result: `SELECT q.remote_id, d.*` where the `document` side may be entirely NULL, decoded as either a full `DocumentRecord` or a skeleton carrying only `remote_id`. Through associations that becomes a nested optional-row decode which is harder to follow than the SQL, for no gain.

The second reason to keep it as one string is that `queryWindowSQL` is shared verbatim by the one-shot read and the `ValueObservation`. Having exactly one definition of the windowed replay — rather than a query built twice and hopefully identically — is worth more here than uniformity with the rest of the file.

I did add a comment making the LEFT-JOIN-yields-skeleton contract explicit, since that's the non-obvious part.

paulgessinger:

## Review round 3 (2026-08-19)

15 follow-up commits at the end of this PR's range, covering Paul's inline comments plus a fresh review pass over the whole 20-commit diff. Ledger [#656](https://github.com/paulgessinger/swift-paperless/issues/656) items 29-34 record what was deliberately **not** fixed.

### The one that would have shipped broken

**`reconcileSavedViewMembership` was destroying every cached list's sort order** (`878a2dc`). It fed `documentIDs(filter:)` — which deliberately forces `id` ascending, because the delete reconcile consumes it as a *set* — straight into `replaceQueryOrder`, which writes those ids as **positions**. The windowed replay reads `ORDER BY q.position`. So every reconcile pass, i.e. every ~5 minutes and on every pull-to-refresh, rewrote the default list and every saved view to document-id order, discarding the sort `fillQuery` had written. The corrective full fill only ran as a periodic backstop, so the wrong order was what the user actually saw.

Fixed with `orderedDocumentIDs(filter:)`, which keeps the query's own sort and appends `id` only as a paging tiebreak. That tiebreak also closes a latent version of the same bug in `fillQuery` and in the R3δ delta, both of which were offset-paging non-unique orderings.

This changes ledger item 20's premise, which is rewritten there: the sweep wasn't failing to *repair* ordering, it was corrupting it. What remains of item 20 is only that `.recentlyBrowsed` has no ordering repair at all.

### Other correctness fixes

| Commit | What it fixes |
|---|---|
| `228ca77` | A fill where **every** view failed still stamped the coverage marker — so a Wi-Fi-but-server-unreachable pass cached nothing, suppressed all retries for the whole window, and showed "last full sync: just now" beside "cached documents: 0". Backstop cadence also 7d → 24h. |
| `2f09791` | The downgrade reclaim was three separate transactions in a fire-and-forget detached task, destructive half first. Backgrounding the app midway left saved views untracked *and* every document on disk, with nothing scheduled to finish. Now one transaction, clears the coverage marker, asks for confirmation first, and won't run at all if the setting write failed. |
| `6b75d4f` | A new server under 2,000 documents was silently stored as *Entire library* and started downloading before the user had seen the setting exists. Now opt-in; `0809d27` turns the size heuristic into a recommendation instead. |
| `712a097` | The detail fill took `prefix(500)` over an unordered query, so a permanently-failing document blocked the head forever — a user without note permission burned 500 doomed requests every foreground and never cached a note. Notes now gated on the permission, failures skipped, and the three whole-table JSON decodes per foreground (one holding the writer lock) are SQL predicates. |
| `d77e715` | `V6` rebuilds `query_order`, and the schema tests only ever ran it against an **empty** database — the row-copy step moved zero rows in every test. Two constraint losses already had to be repaired by follow-ups on this PR; these tests migrate to `V5`, insert rows, then finish, covering the copy, the recreated index, both conflict clauses and the dropped FK. |
| `bcd82f7` | The transfer meter counted decoded `Data.count`, overstating gzipped JSON several-fold and counting `URLCache` hits as if fetched. Now wire bytes from `URLSessionTaskMetrics`. |

### `full_perms`, which is not what the comment said

`Endpoint.documents` requested `full_perms=true` unconditionally. Two problems, both verified against paperless-ngx source rather than inferred:

- It is an **exchange**, not an addition — `OwnedObjectSerializer.__init__` pops `user_can_change` when `full_perms` is set. So `ApiDocument`'s `user_can_change ?? true` fallback fires for every list row and writes a fabricated `true` to disk. Nothing reads it (`DocumentStore.userCanChange` derives from `permissions`), but the comment claimed the opposite of the server's behaviour. Corrected; ledger item 30 covers removing the dead field.
- It is **not free**. On `main` the list action batch-prefetches permissions, but no released version does — I checked `v2.14.7`, `v2.15.3`, `v2.17.1`, `v2.18.4`, and all run `get_permissions(obj)` per object at four guardian queries each. At `pageSize` 250 that's ~1000 extra permission lookups per page, on every list load for every user.

`full_perms` is now a parameter and the envelope-only probes opt out — but the bulk of that cost is **not** gateable without a decision, because `fillQuery` and the interactive list are the same call path and both write to the cache. Recorded as ledger item 29 rather than decided unilaterally.

### Answering the review summary

- **Comments more concise** — `791c7fc`. The three longest blocks roughly halved; `CachingRepository` from 25% to 23% comment lines. Kept every "why", cut restatement of the code.
- **Progress from the fill and reconcile passes** — `bc19400`, and it made the cap removals possible rather than the other way round. New `SyncActivity` (stage, optional saved-view name, completed/total) reported by `fillLibrary`, `fillDocumentDetails` and the delta; the screen shows a stage label and a determinate bar. `isFillingLibrary` is now derived from it, which incidentally fixes two overlapping fills having the first to finish report "Idle".
- **Verifying the skeleton presentation** — enable *Entire library*, let the fill start, force-quit mid-fill, relaunch in airplane mode: membership is written per page but later pages' objects aren't, so the list interleaves loaded and skeleton rows. That mixed state is also where `3d2e6a7` matters — skeleton rows were missing the `listRowInsets` the loaded rows set, so they rendered visibly narrower than their neighbours. I have not driven this in the simulator; happy to if you want a screenshot.

### One thing that moved in your PR

**#627's `V8_AddLastSyncAt` is now `V9`.** This PR needed a `server` column for the per-server cellular toggle, and migration numbering has to match the order they actually run in. Renumbered while resolving, with the registration and the `Stage 11` doc edit carried across.

A test also caught a latent hazard while adding that column: `V2_ImportLegacyConnections` inserted through the *current* `ConnectionRecord` struct, so the moment any later migration adds a column, the V2 data migration starts naming a column that doesn't exist yet and fails for every user upgrading from before it. It now names V1's columns explicitly.

### Verification

Both packages and app built and tested at this PR's head **and** at the stack tip: Common 61, DataModel 230, Networking 217, Persistence 134 — all passing, app builds with zero warnings. No commit in the stack is left conflicted. Nothing was driven in the simulator.

AppShared/Sources/AppShared/Document/DocumentStore.swift
paulgessinger: Does this cover document content itself + files? They probably have different bandwidth metering requirements.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: `json_extract` over basic SQLite test JSON columns are not ideal either?

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: We can't use `JSONB` columns since we target iOS17. How expensive is this now? Maybe we should pull the notes count into an actual column then?

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -79,9 +79,23 @@ public final class DocumentStore: Sendable {
...
+  /// Drives the stage label and progress bar on the Offline & Sync screen.
+  public private(set) var syncActivity: SyncActivity?
+
   /// True while a proactive full-library fill is running (drives the status UI).
-  /// Only the foreground trigger sets this.
-  public private(set) var isFillingLibrary = false
+  /// Derived from ``syncActivity`` rather than a separate flag, so two
+  /// overlapping fills can't have the first one to finish report "idle" while
+  /// the second is still running.
+  public var isFillingLibrary: Bool { syncActivity?.stage == .libraryFill }
```
paulgessinger: Where/how is this used? Maybe worth removing the duplication?

AppShared/Sources/AppShared/Networking/CachingRepository.swift
```diff
@@ -342,17 +369,54 @@ public final class CachingRepository<Wrapped: Repository>: Repository, CachingBa
     // still-missing set simply shrinks and the next pass resumes.
     var fetchedMetadata = 0
     var fetchedNotes = 0
+    defer { progress?(nil) }
+
+    // Notes are a separate resource with their own permission. Without this the
+    // fill asks for every noted document's `/notes/` and takes a 403 each time —
+    // one doomed request per document, on every foreground, forever. `nil` (no
+    // cached matrix yet) stays optimistic, matching `syncElements`.
+    let permissions = try? database.uiSettings(serverID: serverID)?.permissions
```
paulgessinger, line 430: We cannot ask the store here, because the store uses this object, right? Problem is: we then need to ensure that the permissions are up to date from remote, do we not?

paulgessinger:

## Round 4 — answers to the four open threads

I couldn't post these as inline replies: there's an **unsubmitted draft review** on this PR, and GitHub's reply API refuses while one is pending (`user_id can only have one pending review per pull request`). I've left the draft alone. Submit or discard it and I'll thread these properly.

Two commits: `ff755ea` and `89412a0`.

---

### `json_extract` over a TEXT column — you were right, and I under-sold it

Benchmarked before touching anything. 20k documents, realistic payloads including `permissions` (~430 bytes each), SQLite 3.53 on this Mac:

| predicate | as written | real column |
|---|---|---|
| documents needing a `/notes/` fetch | **14.7 ms** | **1.6 ms** (indexed) |
| documents missing file metadata | **19.5 ms** | **6.5 ms** |

~35 ms per foreground on the main actor — and that's a desktop with a much newer SQLite than iOS 17 ships, so on device it'll be several times that. `json_extract` can't use an index, so it's a full scan parsing each row's JSON. The versions predicate is worse: `$.versions` sorts last under `.sortedKeys`, so `json_each` walks past `permissions` in every row before finding anything.

Still much better than the `fetchAll`-and-decode-everything it replaced, but that's not the bar.

**`ff755ea` promotes `notes_count` and `current_version_id` to real columns** (`V9`), with an index on `(server_id, notes_count)` — exactly what `V4` did for `asn`, for the reason its own comment gives. The only JSON path left is the one-time backfill inside the migration.

You're also right that `JSONB` is out: it needs SQLite 3.45 (Jan 2024), iOS 17 ships 3.43.

Two things worth knowing:

- **The backfill and the write path must agree.** One reproduces `Document.currentVersionID` in SQL, the other calls it. There's a test asserting an upgraded row and a freshly written one land on the same value — if they diverged, an upgraded install and a clean one would disagree about which documents still need a metadata fetch.
- I checked whether a missing `notesCount` key could silently drop rows — `json_extract` returns NULL and `NULL > 0` is falsy, where the old Swift decode would have thrown loudly. It can't: `notesCount` is already on `main`, so every cached row has it. Moot now anyway with a `NOT NULL DEFAULT 0` column.

### Cellular gating and document content — no, and they are different problems

The gate covers only the two proactive JSON sweeps, `fillLibraryIfEnabled` and `fillDocumentDetailsIfEnabled`. That's everything this PR downloads on its own initiative.

Document **files** aren't gated because nothing fetches them proactively — `download(document:)` runs when the user opens a document, and gating that would break opening a document on cellular. So "Sync over cellular" honestly means *sync metadata over cellular* today.

The gap worth naming is **thumbnails**: `DocumentListViewModel.prefetchThumbnails` fires a Nuke prefetch every time the observed prefix grows, so scrolling a long list pulls images automatically — not per-image user intent, not covered by this toggle, not trivial on a large library. Pre-existing, but it's the one place the app already spends bandwidth unasked.

Your framing becomes load-bearing the moment anything proactively downloads files (offline-first PDFs being the obvious next step) — that wants its own setting, since the volumes differ by orders of magnitude. **Want the thumbnail prefetch as its own ledger item, or folded under this toggle now?**

### `isFillingLibrary` — right, it was duplication, and the remaining call site was wrong

One usage in the whole stack: `.disabled(store.isFillingLibrary || store.isRefreshing)` on *Sync now*. And it was the wrong predicate — during the **detail fill** the stage is `.detailFill`, so `isFillingLibrary` read `false` and the button re-enabled mid-pass, letting a second pass stack on a running one.

`89412a0` removes the property and disables on `store.syncActivity != nil || store.isRefreshing`, which covers every sweep.

### Migration renumbering, again

`V9` was needed for the column promotion, so **#627's `V9_AddLastSyncAt` is now `V10`**. Second renumber — sorry. The number has to match the order they actually run in, and #617 sits below.

### Verification

Common 61, DataModel 230, Networking 217, Persistence 137 — all passing at this PR's head and at the stack tip; app builds with zero warnings; no commit in the stack is conflicted. The benchmark above is a standalone SQLite harness, not the app, so treat the absolute numbers as indicative and the ratios as the point.

AppShared/Sources/AppShared/Views/Settings/OfflineSyncView.swift
```diff
@@ -32,22 +57,93 @@ public struct OfflineSyncView: View {
...
+              // first; until then the picker snaps back, because `get:` still
+              // reads the unchanged stored mode.
+              if connectionManager.activeOfflineBrowsingMode == .entireLibrary,
+                newMode == .recentlyBrowsed
+              {
+                downgradeRequested = true
+              } else {
+                connectionManager.setOfflineBrowsingMode(newMode)
+                if newMode == .entireLibrary {
+                  Task { await store.fillLibraryIfEnabled(unmetered: unmetered, force: true) }
```
paulgessinger, line 72: This doesn't attach to the view and continues even if I navigate away?

AppShared/Sources/AppShared/Views/Settings/OfflineSyncView.swift
```diff
@@ -113,32 +224,49 @@ public struct OfflineSyncView: View {
         Text(.settings(.offlineSyncDataSince(formattedDate(stats.since))))
       }
     }
+    .task {
+      // The default list's cached status already carries the server's total, so
+      // the size hint costs no request. Absent until that list has been fetched
+      // once, in which case no recommendation is made.
+      guard let key = store.documentQueryKey(filter: .default) else { return }
```
paulgessinger: Might be useful as a helper on store?

Networking/Sources/Networking/Api/Wire/ApiDocument.swift
```diff
@@ -63,6 +63,12 @@ extension ApiDocument {
       notes: notes?.domain ?? NotesPayload(),
       customFields: custom_fields ?? CustomFieldRawEntryList(),
       versions: versions?.map(\.domain) ?? [],
+      // NOTE: not "the server said yes". `full_perms=true` makes paperless-ngx
+      // *drop* `user_can_change` (it swaps the field for `permissions`), and the
+      // document list always asks for full perms — so this fallback fires for
+      // every list-sourced row. Nothing reads `Document.userCanChange`; the real
+      // check is `DocumentStore.userCanChange(document:)`, which derives from
+      // `permissions` + owner and works offline.
       userCanChange: user_can_change ?? true
```
paulgessinger: So should we just drop it?

Networking/Sources/Networking/Api/Endpoint.swift
```diff
@@ -85,18 +85,29 @@ extension Endpoint {
...
+      page: page, rules: filter.rules(for: searchApi), pageSize: pageSize, fields: fields,
+      fullPerms: fullPerms)
 
     var ordering: String = filter.sortField.rawValue
     if filter.sortOrder.reverse {
       ordering = "-" + ordering
     }
+    if stableOrdering, filter.sortField != .other("id") {
+      ordering += ",id"
+    }
```
paulgessinger: Could use an `OrderedSet` from swift-collections + a join

Persistence/Sources/Persistence/Migrations/V2_ImportLegacyConnections.swift
```diff
@@ -56,7 +56,28 @@ enum V2_ImportLegacyConnections {
     }
     for (uuid, legacy) in decoded {
       let record = ConnectionRecord(legacy: legacy, fallbackId: uuid)
-      try record.insert(db)
```
paulgessinger, line 59: This will be an ongoing issue with all the record types right? What's the correct pattern for this? Best practices? Versioned record structs?

Persistence/Sources/Persistence/Migrations/V9_PromoteDocumentQueryColumns.swift
```diff
@@ -0,0 +1,51 @@
...
+/// columns the same queries are ~1.6 ms and ~6.5 ms.
+///
+/// `JSONB` would have made the blob itself cheap to query, but it needs SQLite
+/// 3.45 and iOS 17 ships 3.43.
+///
+/// **Invariant:** both columns are derived from `payload` and written only by
+/// `DocumentRecord.init(serverId:domain:)`, which is the single place a row is
+/// constructed. They are a denormalisation for querying, never an independent
+/// source of truth — `domain` still reads notes count and versions from the
+/// payload.
```
paulgessinger: But maybe it shouldn't?

AppShared/Sources/AppShared/Document/DocumentStore.swift

You were right on both counts — done in `89412a0`.

One usage in the whole stack: `.disabled(store.isFillingLibrary || store.isRefreshing)` on *Sync now*. And it was the **wrong** predicate — during the detail fill the stage is `.detailFill`, so `isFillingLibrary` read `false`, the button re-enabled mid-pass, and a second pass could be stacked on a running one.

Removed the property; the button now disables on `store.syncActivity != nil || store.isRefreshing`, which covers every sweep. `SyncActivity` is the single source for "is something running", so there's nothing left to keep in step.

AppShared/Sources/AppShared/Networking/CachingRepository.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Right on the first, and the second is handled by ordering rather than by anything here.

**Can't ask the store:** correct — `DocumentStore` owns the repository (`store.repository`), so reaching back up would be a cycle. Hence the cached `ui_settings` read.

**Freshness:** the matrix is refreshed immediately before this runs, on the same path. Both lifecycle triggers do `sync()` → `fillLibraryIfEnabled` → `fillDocumentDetailsIfEnabled`, and `sync()` calls `syncElements()`, whose *first* step is `syncUISettings()` — network fetch, write-through to `ui_settings`, and it's already the gate for which element collections get fetched at all. So by the time the detail fill reads permissions they were written seconds earlier, by the same sweep.

Three cases worth stating:

- **Online, permission revoked server-side** — `syncUISettings` picks it up before the fill reads it. Correct.
- **Offline** — `syncUISettings` falls back to the last cached matrix. We proceed on stale-but-best-known permissions, which is the same contract the rest of the offline stack runs on.
- **No matrix at all** (cold start, `ui_settings` never succeeded) — `nil`, and the gate stays optimistic (`?? true`), matching what `syncElements` already does for the collections. We'd rather attempt and take a 403 than silently skip notes for a user who does have permission.

The residual is narrow: permissions revoked *and* the sync failing, so we act on a stale permissive matrix and make doomed requests. The per-session failure set now absorbs that — each id is attempted once, then skipped for the rest of the session.

AppShared/Sources/AppShared/Views/Settings/OfflineSyncView.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Correct — it's unstructured, not view-bound, and it keeps running.

That part is deliberate. The user just switched the setting to *Entire library*; if closing Settings cancelled the fill, enabling the option and backing out would silently do nothing. Same reasoning as *Sync now*. `.task` would bind it to the view's lifetime, which is exactly what we don't want here.

What's genuinely missing is that **nothing can cancel it at all** — that's ledger [#656](https://github.com/paulgessinger/swift-paperless/issues/656) item 31. Two parts:

- `DocumentStore` keeps no handle on the fill, so a server switch leaves it paging the *old* server, and a downgrade reclaims rows the still-running fill writes straight back.
- Even with a handle, cancellation wouldn't reach the pager: `fillLibrary` does `await handle.awaitCompletion()`, which is `await fillTask.value` on a `Task.detached` — detached tasks don't inherit cancellation and awaiting a `Task<Void, Never>` doesn't throw on it. It needs `withTaskCancellationHandler { … } onCancel: { handle.cancel() }`.

It's also why rapid toggling can stack fills. I scoped that to #618 because #618 owns the multi-server switch, but it's this PR's code that creates the un-held task, so I'm happy to pull it forward here instead — say the word.

AppShared/Sources/AppShared/Views/Settings/OfflineSyncView.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Agreed — `6c4eaea` moves it to `DocumentStore.libraryDocumentCount()`.

The view is now just `.task { libraryTotal = await store.libraryDocumentCount() }`. The one-shot-read-of-a-live-stream trick and the error suppression sit behind the store's API, where they belong, rather than being open-coded in a settings screen. Its doc comment says why it's one-shot: callers want the size to make a decision, not to track it.

Networking/Sources/Networking/Api/Wire/ApiDocument.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Yes — dropped in `1881e23`.

Removed from the domain type, the wire type and the stored payload. Nothing read it: every production check goes through `DocumentStore.userCanChange(document:)`, which derives from `permissions` + owner and works offline.

The two wire tests that covered it now pin what still matters — a payload that **omits** the field (a `full_perms` response) and one that **carries** it (older backends) both decode cleanly. Old cached rows keep the key in their JSON blob; the decoder ignores it.

That closes ledger item 30.

Networking/Sources/Networking/Api/Endpoint.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Took the shape, not the dependency — `f2f826c`.

`swift-collections` isn't a dependency of any package here, and pulling it in for one query parameter felt like a poor trade. The same semantics are four lines locally: build the terms, dedupe on the *field* (ignoring the `-` direction prefix), join.

```swift
var terms = [filter.sortOrder.reverse ? "-\(field)" : field]
if stableOrdering { terms.append("id") }
var seenFields = Set<String>()
let ordering = terms
  .filter { seenFields.insert($0.hasPrefix("-") ? String($0.dropFirst()) : $0).inserted }
  .joined(separator: ",")
```

Better than what it replaced in one real way: the old special case tested `filter.sortField != .other("id")`, which only caught the ascending spelling. Deduping on the field name handles `-id` too — added a test pinning that. And another tiebreak later needs no new branch.

Happy to swap in `OrderedSet` if you'd rather have the dependency for other reasons.

Persistence/Sources/Persistence/Migrations/V2_ImportLegacyConnections.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Yes, ongoing — and my earlier fix only closed half of it. Fixed properly in `cd93952f`.

**The rule**, and it's GRDB's own, stated directly in its migration docs:

> Migrations and the rest of the application code do not live at the same "moment". Migrations describe the past states of the database, while the rest of the application code targets the latest one only. This difference is the reason why **migrations should not depend on application types.**
>
> […] migrations should talk to the database, only to the database, and use the database language.

So: literal table names, literal column lists, literal SQL. Not `Record.databaseTableName`, not `Columns.foo.name`, and — the part I'd missed — **not the record's JSON coders or nested storage structs either.**

**Versioned record structs: only where a data migration actually writes rows.** For schema migrations, plain SQL is enough and a frozen struct is dead weight. For a data migration the frozen copy has to live *in* the migration, which is what I've now done: `V2` carries its own `V1StoredUser` / `V1StoredHeader` and its own encoder, and builds the row straight from the legacy payload. It no longer mentions `ConnectionRecord` at all.

**Why the half-fix wasn't enough:** I'd pinned the *column list* but still went through `ConnectionRecord(legacy:fallbackId:)` and `ConnectionRecord.storageEncoder`. Same class of bug, one level down — a rename inside `ConnectionRecord.StoredUser` would have silently changed the JSON `V2` writes into the `user` column, and nothing would have failed loudly.

**Audit:** `V2` was the only migration touching application types; the rest were already string-only. I've written the rule into `Migrations.swift`'s doc comment with the concrete incident (`V9`'s new column broke `V2` at launch for anyone upgrading), so the next person meets it before writing a data migration rather than after.

Persistence/Sources/Persistence/Migrations/V9_PromoteDocumentQueryColumns.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Agreed — `8fcee21` makes the column the only home.

`notesCount` is gone from `DocumentRecord.Payload`; `domain` reads the count from the column. One value, one place, nothing to drift.

The two columns aren't the same kind of thing though, and the migration comment now says so:

- **`notes_count` replaces** the payload's copy. Nothing else needs the count, so it's a plain move, not a denormalisation.
- **`current_version_id` genuinely is a denormalisation** — `payload.versions` has to keep the full version objects, so this is an index over them rather than their home. Kept honest by being written only in `init(serverId:domain:)`, right beside the payload it's derived from, and by the test asserting the SQL backfill and the Swift write path agree.

The backfill still reads `$.notesCount` because rows written before the migration have it; rows written after don't carry the key at all.

AppShared/Sources/AppShared/Views/Settings/OfflineSyncView.swift
paulgessinger: 🤖 _Posted by Claude on Paul's behalf._

Done — `9b1d3f9`, pulled forward onto #617.

Three changes:

- **`fillLibrary` bridges cancellation into the pager.** `await handle.awaitCompletion()` is now wrapped in `withTaskCancellationHandler { … } onCancel: { handle.cancel() }`, plus a `checkCancellation()` after it. A cancelled pass now stops *within* the current view instead of paging it to the end first.
- **`fillLibrary` re-checks the browsing mode per view.** A downgrade mid-pass stops it, rather than racing `reclaimAfterDowngrade` and writing back rows it just reclaimed. One connection-row read per view, and there are few views.
- **`DocumentStore` retains the fill task.** A second request joins the in-flight one rather than starting a parallel pass over the same queries; `set(repository:)` and `deinit` cancel it; the completion path only clears the handle if it's still its own, so a cancel-and-restart can't have its task nulled by the old awaiter.

On `force` being lost when joining: it isn't. A fill only runs at all when the coverage marker is stale or `force` was set, so an in-flight one is already doing the work the forced call wanted.

The thing that tipped it from "defer to #618" to "do it now" was `BackgroundSyncCoordinator` ending its run by reading `Task.isCancelled` — #627 is written assuming a cancelled BGTask stops the work, and until this it couldn't. Only the cancel-on-server-switch line properly belongs to #618, and `set(repository:)` covers it here anyway.

Ledger item 31 closed.

Verification: Common 61, DataModel 230, Networking 218, Persistence 137, all passing at this PR's head and at the stack tip; app builds with zero warnings. One conflict, in #618's `Logger.shared` → `Logger.sync` commit, resolved in that commit — the two new log lines are scoped under `Sync` with the rest.

paulgessinger:

## Round 6 — the four follow-ups

Two commits: `cb4804c` and `6863703`.

### The transfer stats were my regression, and worse than "only Total and Other"

Not expected — **everything** was landing in `.other`, so the breakdown was dead, not just sparse.

Cause is the wire-bytes change from round 4. `NetworkTransfer.category` is a `@TaskLocal`, and the old `record(bytes: data.count)` ran inside `fetchData`, i.e. inside the caller's `withValue(.fill) { … }` scope. Moving the count to `URLSessionTaskMetrics` moved it to a **delegate-queue callback**, which is outside that task — so the task-local read its `.other` default every time.

Fixed by capturing the category at request time, where the scope is still live, and passing it as a value into the callback (`record(bytes:category:)`). Four tests pin it, including one that records from a `Task.detached` — the same shape as a delegate callback — and asserts the captured category survives.

Worth saying plainly: this is exactly the failure the round-4 numbers couldn't catch, because the benchmark measured SQLite and the byte accounting was verified only by reading it.

### Activity flicking to Idle between stages

Real, and the cause is worse than gaps. `syncActivity` was one property written by three sweeps, each with its own `defer { … = nil }` — and they **overlap**: `sync()` fires the reconcile off in its own task and returns, so a reconcile is usually still running when the library fill starts. Whichever finished first blanked the screen while the others carried on. That's why it sat on "Idle" for seconds rather than flickering.

The store now keeps one entry per running sweep, keyed by stage, and publishes the most significant (library fill > detail fill > reconcile > element sync). A sweep reporting "done" retires only its own stage. Both fills also claim their stage *before* their setup reads, so the seed and the two "what's still missing" queries at the top of the detail fill aren't a gap either.

### Progress in the Activity row, and the short divider

Done — and yes, folding it in is what fixes the divider. They're now one cell: a `VStack` holding the `LabeledContent` with the bar and caption beneath it. As a separate row the progress block carried its own separator, which is what wasn't running full width. No layout breakage from putting a `ProgressView` under `LabeledContent` in a `Form` row.

One small improvement while there: `ProgressView(value: activity.fraction)` takes the optional directly, so a stage that hasn't worked out its total yet gets an indeterminate bar instead of a separate branch.

### Verification

Common 61, DataModel 230, Networking 222, Persistence 137 — all passing at this PR's head and at the stack tip; app builds with zero warnings. One conflict in #618's `Logger.sync` commit, resolved there.

**Caveat on the two visual items:** I can't reach the Offline & Sync screen in the simulator without a server, so the row merge and the divider are build-verified only. The reasoning is sound and the structure is straightforward, but they want your eyes — the transfer-stats fix is the one I'd trust from the tests alone.

paulgessinger:

## Round 7 — all running stages, and element sync among them

`a8cbcc1` and `724c608`.

### Showing every stage

Publishing one "most significant" activity was still lossy for exactly the reason the last round diagnosed: the stages overlap, so naming one hid work genuinely in progress. The store now publishes `syncActivities` — every running stage with its own progress — and the screen renders a row each.

Ordering is **fixed** (element sync → library fill → detail fill → reconcile), not by recency or start time. The store holds these in a dictionary, so the order it hands over is arbitrary, and a list that reshuffles as each stage reports progress is unreadable. `SyncActivity` is `Identifiable` by stage, so a row keeps its identity across progress updates instead of being torn down and rebuilt.

### On the disclosure arrow — I didn't add one, and I think that's right

Once element sync reports (below), there are four possible stages and realistically **two at once**: a reconcile plus one fill. Two rows don't need collapsing behind an affordance, and hiding in-progress work behind one is precisely what this change is undoing. Easy to add later if the count ever grows; say the word if you'd rather have it now regardless.

### Element sync — yes, and it was the missing half

You were right to push on this. `.elementSync` was a stage with *no producer*: element sync surfaced only through the `isRefreshing` boolean, which could say "something is happening" and nothing else. I'd actually deleted the case in `a8cbcc1` for being dead — giving it a real producer is the better answer, so it's back in `724c608`.

`syncElements` fans its collections out through a task group, so progress is the group's completion count. Two details worth noting:

- The total is accumulated **as the group is built**, so it's final before the first completion is awaited.
- It counts only collections the user can actually view. A restricted account gets a bar that reaches the end, instead of a nominal eight-of-eight it can never hit.

That makes `isRefreshing` redundant rather than complementary — its only consumer in the whole stack was this screen, and the stage now says the same thing with a count attached. Removed rather than left as a second, coarser answer to the same question.

### Tests

`SyncActivity` moved to `DataModel`, for the same reason `OfflineLibrarySize` lives there: the ordering rule is easy to regress silently and `AppShared` has no test target. Six tests cover it — fixed order regardless of input order, coverage of every stage (so a new one can't silently sort first), the overlapping pair that actually occurs, fraction clamping, and stage-as-identity.

### Verification

Common 61, DataModel 236, Networking 222, Persistence 137 — all passing at this PR's head and at the stack tip; app builds with zero warnings. One conflict in #627's activity-label commit, resolved there so the background-run branch survives alongside the new shape.

Still build-verified only for the visual result — I can't reach this screen in the simulator without a server.

PR #620, opened 2026-07-12
fix: show cached PDF immediately while document metadata refreshes

paulgessinger:
This pull request significantly improves the document detail view's refresh and PDF loading behavior, especially around picking up new versions and providing immediate feedback to users. The main changes ensure that cached PDFs are shown instantly, refreshing an open document updates all relevant data, and iPad users can now refresh from the toolbar. Additionally, error handling is improved to avoid duplicate error toasts and to ensure that background tasks are properly managed.

**Document loading and refresh improvements:**

* Cached PDFs are now displayed immediately when opening a document, even on slow connections, instead of waiting behind a loading spinner.
* Pull-to-refresh and toolbar refresh now pick up new PDF versions, notes, and metadata from the server without requiring the user to close and reopen the document. ( [[1]](diffhunk://#diff-1d9d269fef243f887f3cba163da17436ab549f448550927d7e82f01ae6695138L668-R669) [[2]](diffhunk://#diff-1d9d269fef243f887f3cba163da17436ab549f448550927d7e82f01ae6695138R756-R779)
* On iPad, a new toolbar button allows users to refresh an open document, with an accessible label and progress indicator. [[1]](diffhunk://#diff-436ebe670cf1bc8e284b340bcb6e6b490b0a1b5ca655420c0f2d91b0fa1a7e07R5386-R5397) [[2]](diffhunk://#diff-1d9d269fef243f887f3cba163da17436ab549f448550927d7e82f01ae6695138R756-R779)

**Error handling and task management:**

* Errors during refresh are now deduplicated so that only distinct failures show toasts, preventing multiple identical error messages during a server outage. ([swift-paperless/ViewModel/DocumentDetailModel.swiftR23-R65](diffhunk://#diff-a25456c60451be78aec488c706d66c1519b51117b9d340fd2b14daf4b85e0190R23-R65), Fa366385L158R158)
* The document loading process is now managed by the view model, ensuring that background tasks are properly cancelled when the view disappears, preventing stray toasts or wasted work. [[1]](diffhunk://#diff-a25456c60451be78aec488c706d66c1519b51117b9d340fd2b14daf4b85e0190R93-R97) [[2]](diffhunk://#diff-a25456c60451be78aec488c706d66c1519b51117b9d340fd2b14daf4b85e0190R120-R151) [[3]](diffhunk://#diff-1d9d269fef243f887f3cba163da17436ab549f448550927d7e82f01ae6695138L791-R823)

**Technical and code quality improvements:**

* The Equatable implementation for `DocumentDownloadState` now compares PDF document identity, ensuring that SwiftUI properly updates the preview when a new PDF is loaded.
* The `PDFKitView` now only updates the displayed PDF when the underlying `PDFDocument` instance actually changes, preserving scroll position and avoiding unnecessary reloads.

**Changelog and localization:**

* The changelog is updated to reflect these improvements.
* A new localization string is added for the iPad refresh toolbar button.

### DIFF
--- a/AppShared/Sources/AppShared/Resources/Localization/App.xcstrings
+++ b/AppShared/Sources/AppShared/Resources/Localization/App.xcstrings
@@ -5383,6 +5383,18 @@
         }
       }
     },
+    "documentDetailRefresh" : {
+      "comment" : "Toolbar button on the iPad document detail view that reloads the document from the server.",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Refresh"
+          }
+        }
+      }
+    },
     "documentDuplicateAsn" : {
       "extractionState" : "manual",
       "localizations" : {
--- a/AppShared/Sources/AppShared/Views/PDFKitView.swift
+++ b/AppShared/Sources/AppShared/Views/PDFKitView.swift
@@ -51,7 +51,24 @@ public struct PDFKitView: UIViewRepresentable {
     self.pageIndex = pageIndex
   }
 
-  public func updateUIView(_: PDFKit.PDFView, context _: Context) {}
+  // PDFView keeps whatever document `makeUIView` gave it, so without this a
+  // SwiftUI update that swaps in a different `PDFDocument` at the same tree
+  // position — a re-download after a server-side version bump, or the
+  // archived/original switch — silently keeps rendering the old file.
+  //
+  // Identity, not equality: `PDFDocument` has no meaningful `==`, and
+  // `PDFDocument.loadBackground` always constructs a fresh instance, so `!==`
+  // is precisely "this is a different load". Guarding on it also keeps the
+  // update a no-op for the unrelated passes (scrolling, layout, size-class
+  // changes) that would otherwise reassign the document — which resets the
+  // user's scroll position — on every frame.
+  public func updateUIView(_ uiView: PDFKit.PDFView, context _: Context) {
+    guard uiView.document !== document else { return }
+    uiView.document = document
+    if let pageIndex, let page = document.page(at: pageIndex) {
+      uiView.go(to: page)
+    }
+  }
 
   public func makeUIView(context _: Context) -> PDFKit.PDFView {
     let view = PDFKit.PDFView()
--- a/current_changelog.txt
+++ b/current_changelog.txt
@@ -29,3 +29,4 @@
 - Opening a document whose PDF is already cached now shows it immediately, even on a slow connection, instead of waiting behind a loading spinner
 - Pulling to refresh an open document now picks up a new version added on the server, instead of continuing to show the previous one until you close and reopen it
 - Pulling to refresh a document that has a new version on the server now finishes refreshing its notes, metadata and suggestions, instead of stopping partway and showing an "Unexpected error"
+- On iPad you can now refresh an open document from the toolbar, picking up a new version or changed details without closing and reopening it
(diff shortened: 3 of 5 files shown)


paulgessinger:

One item from [#600's review](https://github.com/paulgessinger/swift-paperless/pull/600#issuecomment-5306727572) belongs here — ledger #656, item 22. **Not implemented.**

**The notes prefetch toasts a permissions error on every refresh.** `DocumentDetailModel.loadNotes` routes through `store.notes(for:)`, which opens with `try checkPermission(.view, for: .note)` and throws `PermissionsError` — a `DisplayableError` — straight into `onError`. `DocumentNoteView` guards with `canViewNotes` before loading; this call site doesn't. A user without the note view permission gets "You don't have permission to view notes" on every detail pull-to-refresh, on every document.

Two things worth knowing while fixing it:

- **This PR's error collapse doesn't absorb it.** `PermissionsError` differs from every other error in the pass, so it survives the dedup as one extra spurious toast.
- **The method's doc comment asserts the opposite of the behaviour** — "Routed through `store.notes(for:)` so the `.note` view-permission gate is respected". Respected here means *throws*. Worth fixing the comment with the code, since it currently reads as though the case were handled.

One line: gate on `store.permissions.test(.view, for: .note)`, mirroring `canViewNotes`.

While verifying this, one thing about this PR's own fix that's worth recording: `1a97237c4`'s collector dedups with `==`, and a bridged `URLError` carries a per-request task id in `userInfo`, so no two ever compare equal. The collapse only actually works once `10021dc3` on #627 normalizes transport failures into the `Equatable` `RequestError`. At this PR's head it's nominal.

paulgessinger:

## Manual test still owed: refresh cancellation **without** a new document version

`859716b1b fix: don't let a pull-to-refresh cancel its own document load` needs an explicit
re-test on a document that has **no new version on the server**, before this is considered
verified.

**Why.** Testing #600's head on device, pull-to-refresh inside the document detail view
produced a "cancelled" error on **every** refresh, with nothing changing during the refresh
and no new version involved. That commit's message frames the trigger narrowly:

> `loadDocument` runs first and, on a version bump, swaps in the new PDF — replacing the
> preview inside that very scroll view. The refresh session ends, and the
> metadata/notes/suggestions legs still in flight are cancelled.

The observed behaviour is broader than that description: it is not limited to a version bump.

**Why it should still be fixed anyway.** The fix is not trigger-specific. It swaps
`model.load(...)` for `model.startLoad(...)` in `.refreshable`, moving the work into an
unstructured `Task` that does not inherit the gesture task's cancellation. Whatever ends the
refresh session — version swap, PDF re-render, any content-height change, or something else
entirely — no longer kills the in-flight legs. So the no-version case should be covered by
the same change.

**But that is reasoning, not evidence.** The fix was developed and written up against the
version-bump repro, so the broader case has never actually been exercised against it.

### What to check

- [x] Open a document with **no** new version on the server. Pull to refresh repeatedly (5–10×). No "cancelled" toast, no "Unexpected error".
- [x] Confirm the refresh actually *completes* rather than merely failing silently — notes, metadata and suggestions should all refresh, not just the ones with a cache fallback. (In the original bug only the legs without a fallback surfaced, which is why it appeared as two errors rather than four.)
- [x] Re-check the version-bump case too, so the original repro is not regressed by anything since.
- [x] Dismiss the view mid-refresh and confirm nothing toasts over the screen that replaces it — `onDisappear` → `cancelLoad()` is meant to keep that boundary.

Recorded here rather than on #600, since that is where the bug is live and this is where the
fix lives.

paulgessinger:
The same-version refresh path can still leave the rendered PDF stale.

When `modified` changes but `currentVersionID` does not, `ContentStore.store` replaces the cached archive at the same canonical URL. The URL equality check in `runDownload` then returns before constructing a new `PDFDocument`, so the existing PDFKit view keeps rendering the old document instance. This affects regenerated archives and older servers without document-version data.

The post-metadata download needs to reload the PDF when the document metadata changed even if the returned URL matches the displayed file—for example, by passing a force-reload flag for that path.

PR #627, opened 2026-07-17
feat: Background sync

paulgessinger:
This pull request introduces Stage 11 background sync support, enabling the app to perform background synchronization tasks reliably, even when the UI is not active. It also improves error messaging for connectivity issues and refines how network status is detected for background tasks. The main changes are grouped as follows:

**Background Sync Infrastructure:**

* Added a new `BackgroundSyncCoordinator` class to coordinate background sync tasks, ensuring only one `CachingRepository` per server runs at a time and supporting both UI-driven and headless (background-only) execution. (`AppShared/Sources/AppShared/Networking/BackgroundSyncCoordinator.swift`)
* Introduced a `NetworkPathProbe` utility that reads the raw path cost (`isExpensive` / `isConstrained`) for background runs. The heavy-fill decision is not made here: each server combines that cost with its own `syncOverCellular` via `SyncCondition`, so a heavy sync runs only where that particular server allows it. (`AppShared/Sources/AppShared/Networking/NetworkPathProbe.swift`)
* Updated `SyncEngine` to support different sync "tiers" (cheap vs. full), background-safe server selection, and stalest-first sweep ordering. (`AppShared/Sources/AppShared/Networking/SyncEngine.swift`) [[1]](diffhunk://#diff-5682fab0cb4c34ee6b21447d47f8cbe3495cfb9fa125780f8bdf514b93b129c7L13-R26) [[2]](diffhunk://#diff-5682fab0cb4c34ee6b21447d47f8cbe3495cfb9fa125780f8bdf514b93b129c7R39-R58) [[3]](diffhunk://#diff-5682fab0cb4c34ee6b21447d47f8cbe3495cfb9fa125780f8bdf514b93b129c7L102-R146)

**Sync State Tracking and UI Integration:**

[...]

PR #685, opened 2026-08-23
refactor: App-level sync orchestration

paulgessinger:
This pull request refactors the `DocumentStore` to delegate server ownership and sync responsibilities to a `ServerSession`, improving separation of concerns and preventing multiple owners from driving the same server. The store now activates and syncs through a session, rather than directly managing repositories and sync tasks. This enables more robust handling of server switching, progress reporting, and background tasks.

Key changes include:

**Session Ownership and Activation:**

* `DocumentStore` now holds a reference to a `ServerSession` (or a registry to obtain one) instead of directly owning a repository, and exposes the repository via a computed property that reads from the session. The store is constructed either with a session (for previews/tests) or a registry (for production). [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L148-R167) [[2]](diffhunk://#diff-cfd0bb6a054df08870aa9bda104e7cc9b2204aba2ebb6204ba5340e821a6c725L51-R54)
* The `activate(connection:)` method now obtains and prepares a session from the registry, and swaps the session via a new `install(session:reload:)` method, rather than swapping repositories directly.

**Sync and Progress Reporting:**

* Sync activities (`syncActivities`, `isSyncing`, `lastReconcileAt`) are now projections from the session, not state managed by the store. This ensures the UI reflects background or scheduler-initiated syncs, not just those initiated by the store.
* The element sync and reconcile logic now runs through the session (`session.syncElements()`, `session.reconcileDocuments(force:)`), and the previous coalescing/cancellation logic in the store is removed. [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L480-R518) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L523-L554)

**Repository and Backend Access:**

* All code paths that previously downcast the store's repository to a `CachingBackend` now access the backend via the session (`session?.backend`). [[1]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L214-R220) [[2]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L815-R801) [[3]](diffhunk://#diff-2f7b34a418ea1f6972b552ed9fe8da0466ab897fa87ef7940f2fc850a7d84f70L834-R813)

**Lifecycle and Memory Management:**

* The store no longer cancels in-flight sync or fill tasks when switching servers; these tasks are now owned by the outgoing session and allowed to complete. This prevents wasted work and improves consistency when switching servers mid-operation.

**Preview and Test Support:**

* The preview initializer for `DocumentStore` is updated to use a session, matching the production code path and ensuring consistent ownership semantics.

These changes collectively improve the reliability and maintainability of server activation, sync, and progress reporting in the app.

**Stack** (bottom to top):

<details>

</details>

- #618 feat/multi-server-sync
- #685 refactor/server-session-ownership  👈
- #627 feat/background-sync

### DIFF
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -102,10 +102,11 @@ public final class DocumentStore: Sendable {
   /// When the document reconcile sweep (R2/R3δ/membership) last **succeeded**.
   /// `nil` until the first successful reconcile this session.
   ///
-  /// Observed (not `@ObservationIgnored`) because the Offline & Sync screen
-  /// renders it: untracked storage would leave the row reading "Never" until
-  /// some other observed property happened to fire.
-  public private(set) var lastReconcileAt: Date?
+  /// Owned by the session, not mirrored into stored state here: the session's
+  /// own stamp is observed, so a view reading this through the store still
+  /// repaints when a sweep lands — including a sweep the *scheduler* ran, which
+  /// the store used never to hear about.
+  public var lastReconcileAt: Date? { session?.lastReconcileSuccess }
 
   /// When the active server's library was last fully filled (`nil` if never).
   /// A stored, observed mirror of `server_sync_state.library_coverage_at` —
@@ -175,20 +176,6 @@ public final class DocumentStore: Sendable {
   @ObservationIgnored
   private var taskUpdateTask: Task<Void, Never>?
 
-  // The in-flight element sync, if any. Concurrent `sync` callers join this one
-  // task instead of each firing their own `syncElements` (the launch flurry of
-  // sync / fetchUISettings / scenePhase triggers shares a single network
-  // pass). Only ever touched on the main actor.
-  @ObservationIgnored
-  private var syncTask: Task<Void, Error>?
-
-  // The in-flight proactive library fill. Held so it can be cancelled when the
-  // repository is replaced, and so a second request joins it rather than
-  // starting a parallel pass over the same queries. Only touched on the main
-  // actor.
-  @ObservationIgnored
-  private var libraryFillTask: Task<Void, Never>?
-
   // Observes the active server's `library_coverage_at` into `libraryCoverageAt`.
   // Re-pointed alongside the element projection whenever the repository changes.
   @ObservationIgnored
@@ -201,14 +188,6 @@ public final class DocumentStore: Sendable {
   // Observes the active server's cached document count into `cachedDocumentCount`.
   @ObservationIgnored
   private var documentCountObservationTask: Task<Void, Never>?
-
-  // When the reconcile last *ran*, which is what the throttle needs — it has to
-  // advance on every attempt or a failing server would re-sweep its whole id set
-  // on every foreground. `lastReconcileAt` is the separate, user-facing stamp and
-  // only advances on success.
-  @ObservationIgnored
-  private var lastDocumentReconcile: Date?
-  private let reconcileThrottle: TimeInterval = 300
 
   // MARK: Methods
 
@@ -231,10 +210,12 @@ public final class DocumentStore: Sendable {
     self.session = session
     imagePipeline = Self.makeImagePipeline(delegate: session?.repository?.delegate)
     wireElementStore()
+    session?.progress = { [weak self] activity, stage in
+      self?.report(activity, for: stage)
+    }
   }
 
   deinit {
-    libraryFillTask?.cancel()
     taskUpdateTask?.cancel()
     coverageObservationTask?.cancel()
     syncErrorObservationTask?.cancel()
@@ -366,21 +347,24 @@ public final class DocumentStore: Sendable {
   /// that needs to swap repositories should route through here so the invariant
   /// below keeps covering it.
   private func install(session: ServerSession, reload: Bool) {
-    // The backstop against installing a non-caching repository is gone because
-    // it is now unrepresentable: a session only ever yields a `CachingBackend`.
+    // Nothing is cancelled here any more, and that is the point.
     //
-    // The cancel-on-swap below, however, is still load-bearing *in this commit*.
-    // The store — not the session — still owns the element sync and the fill, so
-    // an in-flight one belongs to the outgoing backend: it writes the old
-    // server's rows, and `runSyncElements` would let the caller's follow-up sync
-    // join it (both `activate` callers do `activate` → `sync()`) and return
-    // without ever fetching the new server's cache. Once the sequence moves onto
-    // the session, that work is simply an *inactive* session's work and should
-    // keep running, and this goes away.
-    syncTask?.cancel()
-    syncTask = nil
-    libraryFillTask?.cancel()
-    libraryFillTask = nil
+    // The store used to cancel the in-flight element sync and library fill on
+    // every swap, because it owned them: they wrote the *outgoing* server's rows
+    // and a follow-up `sync()` would join them and return without ever fetching
+    // the new server's cache. Now that work belongs to the outgoing *session* —
+    // it is simply an inactive server's work, keyed to that server, and there is
+    // no way for the incoming server's sync to join it. So it keeps running, and
+    // switching servers mid-fill no longer throws away the pages already paid
+    // for.
+    //
+    // The progress hook does move, though: the outgoing session must stop
+    // reporting into this store's `syncActivities`, or the Offline & Sync screen
+    // would show the previous server's stages.
+    self.session?.progress = nil
+    session.progress = { [weak self] activity, stage in
+      self?.report(activity, for: stage)
+    }
     self.session = session
     imagePipeline = Self.makeImagePipeline(delegate: session.repository?.delegate)
     wireElementStore()
@@ -528,18 +512,18 @@ public final class DocumentStore: Sendable {
   public func sync(userInitiated: Bool = false) async throws {
     Logger.sync.notice("Sync store (userInitiated: \(userInitiated))")
     do {
-      try await runSyncElements()
+      try await session?.syncElements()
       lastSyncError = nil
       Logger.sync.info("Sync store complete")
       // Reconcile remote deletes alongside the element sync (throttled,
       // non-blocking). Pull-to-refresh (userInitiated) bypasses the throttle.
       let userInitiated = userInitiated
-      Task { [weak self] in await self?.reconcileDocuments(userInitiated: userInitiated) }
+      Task { [weak self] in await self?.session?.reconcileDocuments(force: userInitiated) }
     } catch {
-      // A cancellation is never the user's problem to see: either the caller's
-      // own task went away, or `install(session:)` retired this sync on a
-      // connection switch. Drop it before the userInitiated rethrow so neither
-      // case toasts.
+      // A cancellation is never the user's problem to see: the caller's own task
+      // went away. (A connection switch no longer retires it — that sync belongs
+      // to the outgoing session and runs on.) Drop it before the userInitiated
+      // rethrow so it doesn't toast.
       if error.isCancellationError {
         Logger.sync.debug("Element sync cancelled")
         return
@@ -553,38 +537,6 @@ public final class DocumentStore: Sendable {
       }
       Logger.sync.error("Background sync failed (suppressed): \(error)")
     }
-  }
-
-  /// Run (or join) the single in-flight `syncElements`. Returns when it
-  /// completes; throws its error to every joined caller (who each decide what to
-  /// do with it). A no-op without a caching backend.
-  private func runSyncElements() async throws {
-    if let syncTask {
-      Logger.sync.debug("Joining in-flight element sync")
-      return try await syncTask.value
-    }
-    guard let backend = session?.backend else {
-      // No DB-backed repository (e.g. NullRepository before login). Nothing to
-      // sync; the projection is empty until a caching repository is set.
-      Logger.sync.info("Sync skipped: repository is not a caching backend")
-      return
-    }
-    Logger.sync.debug("Starting element sync")
-    let task = Task { [weak self] in
-      try await NetworkTransfer.$category.withValue(.sync) {
-        try await backend.syncElements { self?.report($0, for: .elementSync) }
-      }
-    }
-    syncTask = task
-    defer {
-      // Retract only what we installed. `install(session:)` can retire this sync
-      // mid-flight and a replacement can already own `syncTask` by the time we
-      // resume; clearing it blindly would break the replacement's coalescing.
-      if syncTask == task {
-        syncTask = nil
-      }
-    }
-    try await task.value
   }
 
   public func document(id: UInt) async throws -> Document? {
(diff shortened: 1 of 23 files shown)


paulgessinger:
@codex

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -493,18 +490,18 @@ public final class DocumentStore: Sendable {
...
     do {
-      try await runSyncElements()
+      try await session?.syncElements()
       lastSyncError = nil
       Logger.sync.info("Sync store complete")
       // Reconcile remote deletes alongside the element sync (throttled,
       // non-blocking). Pull-to-refresh (userInitiated) bypasses the throttle.
       let userInitiated = userInitiated
-      Task { [weak self] in await self?.reconcileDocuments(userInitiated: userInitiated) }
+      Task { [weak self] in await self?.session?.reconcileDocuments(force: userInitiated) }
```
paulgessinger: `reconcileDocuments` is the group of sync sweeps and not just a single one, right?

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -840,127 +805,32 @@ extension DocumentStore {
...
 
-  /// Throttled remote-delete reconcile: drop cached documents that no longer
-  /// exist on the server (so they disappear from every offline list). Soft-fail
-  /// (background); kicked from `sync()`. `userInitiated` bypasses the throttle.
+  /// Throttled remote-delete reconcile on the active server, run by its session:
+  /// drop cached documents that no longer exist on the server (so they disappear
+  /// from every offline list), fold in the changed-metadata delta, then rebuild
+  /// saved-view membership. Soft-fail (background); kicked from `sync()`.
+  /// `userInitiated` bypasses the throttle.
   public func reconcileDocuments(userInitiated: Bool = false) async {
```
paulgessinger, line 831: What do we need the alias for still?

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -840,127 +805,32 @@ extension DocumentStore {
...
+    await session?.reconcileDocuments(force: userInitiated)
   }
 
-  /// Proactive *Entire library* fill, gated by the setting and an unmetered link.
+  /// Proactive *Entire library* fill on the active server, run by its session.
   /// Foreground-only; soft-fail (offline-tolerant). `force` ignores the freshness
   /// marker (e.g. the user just enabled the setting). A no-op when the setting is
-  /// *Recently browsed*, the link is metered, or there's no caching backend.
+  /// *Recently browsed*, the link is metered, or there is no active server.
   public func fillLibraryIfEnabled(unmetered: Bool, force: Bool = false) async {
```
paulgessinger: Same here.

AppShared/Sources/AppShared/Document/DocumentStore.swift
```diff
@@ -840,127 +805,32 @@ extension DocumentStore {
...
   }
 
-  /// Proactive *Entire library* per-document detail fill (notes + file-metadata),
-  /// gated by the setting and an unmetered link. Run after `fillLibraryIfEnabled`
-  /// so the document rows to walk are already on disk. Idempotent and resumable
-  /// (driven off what's still missing), so it needs no `force`. Soft-fail.
+  /// Proactive *Entire library* per-document detail fill (notes + file-metadata)
+  /// on the active server, run by its session. Run after `fillLibraryIfEnabled`
+  /// so the document rows to walk are already on disk. Soft-fail.
   public func fillDocumentDetailsIfEnabled(unmetered: Bool) async {
```
paulgessinger: And here

AppShared/Sources/AppShared/Networking/ServerSession.swift
```diff
@@ -0,0 +1,529 @@
...
+  @ObservationIgnored private var lastReconcileAttempt: Date?
+
+  /// When the reconcile last refreshed *something*. The user-facing "last
+  /// refreshed" stamp the Offline & Sync screen renders, so it is observed.
+  public private(set) var lastReconcileSuccess: Date?
+
+  /// Keeps the active server's on-appear triggers off the wire. Bypassed by
+  /// pull-to-refresh and by the scheduler, which has already applied its own
+  /// (much longer) throttle before asking.
+  @ObservationIgnored private let reconcileThrottle: TimeInterval = 300
```
paulgessinger: Why not static?

AppShared/Sources/AppShared/Networking/ServerSession.swift
```diff
@@ -0,0 +1,529 @@
...
+  // wants the element sync back, not a full server sweep), while the scheduler
+  // runs the composed sequence. Sharing the per-phase tasks is what lets both
+  // callers hit the same session concurrently and coalesce — which in turn is
+  // what let the engine's "skip the active server" guard go away, since there
+  // is no longer a second owner to stay out of the way of.
+  @ObservationIgnored private var syncTask: Task<Void, Never>?
+  @ObservationIgnored private var elementSyncTask: Task<Void, Error>?
+  @ObservationIgnored private var reconcileTask: Task<ReconcileResult, Never>?
+  @ObservationIgnored private var libraryFillTask: Task<Bool, Never>?
+  @ObservationIgnored private var detailFillTask: Task<Bool, Never>?
```
paulgessinger: The stages are ordering dependent though, right? There's nothing structural now that guarantees e.g. the store respects the required ordering, or is ther?

AppShared/Sources/AppShared/Networking/ServerSession.swift
```diff
@@ -0,0 +1,529 @@
...
+  /// store's `sync()` starts the reconcile in its own task and returns, so a
+  /// reconcile is usually still running when the library fill begins. Publishing
+  /// one of them meant whichever finished first blanked the screen to "Idle"
+  /// while the other carried on — and, because progress is reported coarsely, it
+  /// stayed blank until the survivor reached its next checkpoint.
+  ///
+  /// Living on the session rather than the store is what makes an *inactive*
+  /// server's progress observable at all: the engine's sweeps used to report
+  /// nowhere, because the only reporter belonged to the active server's store.
+  public private(set) var syncActivities: [SyncActivity] = []
```
paulgessinger: Naively, this could be a computed property, no?

AppShared/Sources/AppShared/Networking/ServerSession.swift
```diff
@@ -0,0 +1,529 @@
...
+  /// stage; the list keeps showing whatever else is still running.
+  private func report(_ activity: SyncActivity?, for stage: SyncActivity.Stage) {
+    activeStages[stage] = activity
+    syncActivities = activeStages.values.sortedForDisplay
+  }
+
+  public init(
+    serverID: UUID,
+    database: Database,
+    manager: ConnectionManager,
```
paulgessinger: It's a bit awkward that each session has to hold on to the whole connection manager, rather than the relevant connection. What requires this?

AppShared/Sources/AppShared/Networking/ServerSession.swift
```diff
@@ -0,0 +1,529 @@
...
+  // MARK: - Composed sequence
+
+  /// Run (or join) this server's full sync pass.
+  ///
+  /// The sequence is the one the active path has always run — element sync, the
+  /// three reconcile sweeps, then the proactive fills — and is now the only copy
+  /// of it. Nothing throws out: a per-server failure (offline, a rethrown 401)
+  /// is this server's problem alone and must never affect its siblings.
+  ///
+  /// `runHeavyFill` is `SyncPlan`'s decision, not this type's.
```
paulgessinger: I'm a bit fuzzy on what gets decided where? I was trying to steer this to a central place with `SyncCondition`, but we seem to have condition arguments spread all over. Which one does what?

AppShared/Sources/AppShared/Networking/ServerSession.swift
```diff
@@ -0,0 +1,529 @@
...
+    // it blindly would break the replacement's coalescing.
+    if syncTask == task {
+      syncTask = nil
+    }
+  }
+
+  /// Degrade to per-server needs-auth without a network call: deterministic and
+  /// cheap, and the next trigger picks the server up once a token lands. The 401
+  /// path stays a backstop for a token the server rejects.
+  public func markNeedsAuth(_ stored: StoredConnection) {
```
paulgessinger, line 586: Who calls this? Why not directly on manager? Because otherwise we would have to ovserve manager here? I guess that's the design constraint that drives having the manager?

AppShared/Sources/AppShared/Networking/ServerSessionRegistry.swift
```diff
@@ -0,0 +1,132 @@
...
+  @ObservationIgnored private var observationTask: Task<Void, Never>?
+
+  /// Called after each observation tick with the current set and the set as it
+  /// was before, so a scheduler can decide what a newly-appeared server deserves.
+  ///
+  /// The registry deliberately does not make that decision itself: whether a new
+  /// server is synced immediately, and whether the active one is excluded, is
+  /// `SyncPlan`'s call, not lifecycle.
+  @ObservationIgnored
+  public var onServersChanged: (@MainActor (_ current: Set<UUID>, _ previous: Set<UUID>) -> Void)?
```
paulgessinger: Explain this.

AppShared/Sources/AppShared/Networking/SyncEngine.swift
```diff
@@ -30,76 +37,47 @@ import os
...
-          continuation.yield()
-          continuation.finish()
-        }
-        self?.handleConnectionsChanged()
-        for await _ in stream { break }
-      }
+    registry.onServersChanged = { [weak self] current, previous in
+      self?.handleServersChanged(current: current, previous: previous)
     }
+    registry.start()
```
paulgessinger: This is a bit awkward, isn't it? Could it use `@Observation` directly?

paulgessinger (commented):
- So the overall orchestration between `DocumentStore` and the session is that there's a fixed place with task slots, and both the auto-sync and the store-driven active sync sequence using them?

AppShared/Sources/AppShared/Document/DocumentStore.swift
paulgessinger: Yes — all three sweeps: remote deletions, the changed-metadata delta, then the saved-view membership rebuild.

The doc comment on the `DocumentStore` facade was corrected for this in **`refactor: address review nits on the session and registry`**. The comment *at this call site* was missed and still read "Reconcile remote deletes" — fixed now in **`docs: say what the reconcile kicked from sync() actually does`**.

`ServerSession.reconcileDocuments` also spells the three sweeps out, and each carries its own `catch` so one failure doesn't skip the other two.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Document/DocumentStore.swift
paulgessinger: Kept, per your later call — "keep the aliases, probably a good idea to keep session API surface limited".

The alias is doing one real thing beyond forwarding: it maps the store's caller-facing `userInitiated` vocabulary onto the session's `force`, so the ~25 call sites don't have to know that "the user pulled to refresh" and "bypass the 300 s sub-throttle" are the same fact.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Document/DocumentStore.swift
paulgessinger: Kept for the same reason as the `reconcileDocuments` alias — limiting the session's public surface.

This one has since changed shape though: `fillLibraryIfEnabled(unmetered:force:)` and `fillDocumentDetailsIfEnabled(unmetered:)` were collapsed into a single `fillIfEnabled(phases:force:)` in **`refactor: gate the active server's fill on planned phases, not an unmetered flag`**. Two reasons:

- The `unmetered: Bool` was the last survivor of the vocabulary `SyncPhases` replaced — every caller held a raw fact about the link and re-derived the fill rule from it. Callers now ask `SyncPlan.phases(isEntireLibrary:condition:)`, the same function the inactive sweep asks, so the active server can't drift onto a different rule.
- The two fills are one decision (`.fill` covers both) and the details must be walked *after* the pages that put their rows on disk. Every caller already paired them — except the one that didn't, which is why enabling *Entire library* used to leave notes and file metadata until the next launch.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Document/DocumentStore.swift
paulgessinger: Same as above — this one no longer exists as a separate entry point. It folded into `fillIfEnabled(phases:force:)`, which runs the library pages then the details in the required order rather than trusting each caller to pair them.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Networking/ServerSession.swift
paulgessinger: No reason — fixed. It's `private static let reconcileThrottle: TimeInterval = 300` now, in **`refactor: address review nits on the session and registry`**.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Networking/ServerSession.swift
paulgessinger: Good catch, and it drove the biggest change in this round — but I want to be straight that it's **only partly closed**.

What was added (**`refactor: plan emits sync phases instead of a heavy-fill flag`**): a `SyncPhase` enum whose cases are declared in dependency order, and a `SyncPhases` set whose `ordered` property is the single canonical sequence. Callers now choose *which* work to do; the order is not theirs to pick. `SyncServerAction` carries that set instead of a `runHeavyFill: Bool`, so the executor receives a decision rather than the inputs to one.

That makes it structural for the **scheduler's** path: `ServerSession.runSync` walks `phases.ordered` and cannot run a phase before one it depends on.

It is **not** yet structural for the **active** path. `DocumentStore.sync()` still calls `syncElements()` and then kicks `reconcileDocuments` on a detached task, and the app drives the fill separately — so the active server's ordering is still composed by hand, exactly as you suspected. The phase implementations are shared; the composition is not.

Closing that is tracked as follow-up 2 in `plans/sync-followups.md` ("one composed proactive path"). It's deliberately not in this PR: doing it properly also means moving the reconcile off the on-appear trigger, which changes how quickly remotely-deleted documents disappear, and that needs the cost-aware reconcile work first.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Networking/ServerSession.swift
paulgessinger: Yes — changed in **`refactor: address review nits on the session and registry`**:

```swift
public var syncActivities: [SyncActivity] { activeStages.values.sortedForDisplay }
```

over an observed `activeStages` dictionary, so there's no stored mirror to keep in step.

Worth noting the neighbouring `isSyncing` went the *other* way in **`fix: drive the sync-busy state off the session's task slots`** — it has to stay stored, because it's derived from the task slots and `deinit` cancels those, and an observed property can't be read from a nonisolated `deinit`. So the slots stay `@ObservationIgnored` with `didSet { refreshIsSyncing() }`.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Networking/ServerSession.swift
paulgessinger: Nothing required it, in the end — it's gone.

The only thing the session used `ConnectionManager` for was flagging needs-auth, and that is a column on the `server` row (`needs_auth`, present since `V1_CreateServer`). So **`refactor: write needs-auth from the repository layer, not through ConnectionManager`** swapped the manager for a `Database` handle and writes the column directly; `ConnectionManager` rebuilds its in-memory set from its own observer, so it doesn't need telling.

`NeedsAuthRepository` got the same treatment in that commit, which also removed the `manager` parameter from `makeCachingRepository`.

Two holdouts remain — `ServerSessionRegistry` and `SyncEngine` still take a manager for `connections` / `activeConnectionId`. Those are noted in the follow-up plan, because dropping them is also what would let this whole cluster move into a host-testable package.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Networking/ServerSession.swift
paulgessinger: Fair — they were spread out, and this comment is what prompted the redesign. The layering now:

**`LinkCost`** (DataModel) — a fact about the network, nothing else. `isExpensive` + `isConstrained` as one value, published by `NetworkMonitor` and passed unchanged. Previously these travelled as two loose `Bool`s through five layers and were re-packed at every hop, which is how "cost" ended up being a policy in disguise. Added in **`refactor: carry the link's cost as one value instead of two loose flags`**.

**`SyncCondition`** (DataModel) — that fact plus *this server's* `syncOverCellular` opt-in. Separate from `LinkCost` because two servers on the same link can legitimately answer differently.

**`SyncPlan.phases(isEntireLibrary:condition:)`** (DataModel, unit-tested) — the single rule. `.full` when the server is *Entire library* **and** the link allows it, `.cheap` otherwise. Both drivers call it: the sweep via `inactiveActions`, the active server directly. There's a test asserting the sweep plans each server by the same function the active path uses, so they can't drift apart again.

**`ServerSession`** — receives a `SyncPhases` set and executes it. No condition arguments at all any more.

So there is one place the decision is made, one place the fact is measured, and the executor sees neither.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Networking/ServerSession.swift
paulgessinger: `SyncEngine.runAction` calls it, for a server whose `SyncServerAction` came back `needsAuthOnly` — i.e. a configured server with no usable credential yet. It degrades that server without attempting a network call, so an uncredentialed server never fails the sweep and gets picked up on the next trigger once a token lands. The 401 path in `NeedsAuthRepository` stays as the backstop for a token the server *rejects*.

On "why not directly on manager" — that's now moot: it doesn't go through the manager at all. **`refactor: write needs-auth from the repository layer, not through ConnectionManager`** writes `needs_auth` straight to the `server` row via `Database.setNeedsAuth`. `ConnectionManager` hydrates its in-memory set from its own observer, so the write reaches it without the session holding a reference — which removed the design constraint you were pointing at.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Networking/ServerSessionRegistry.swift
paulgessinger: This callback is gone — deleted in **`refactor: address review nits on the session and registry`**.

The registry now publishes its state instead of pushing it: `public private(set) var serverIDs: Set<UUID>`, observed. A scheduler that wants to react to a new server observes it like any other state. The registry still makes no policy decision — whether a new server is synced immediately, and whether the active one is excluded, remains `SyncPlan`'s call.

The reason it was a callback at all was to keep the registry the *sole* observer of the connections table (it maintains a second map keyed by the same UUIDs, and two independent observers would drift). Publishing a set preserves that — the registry is still the only thing watching `manager.connections` — without the inversion of control.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Networking/SyncEngine.swift
paulgessinger: Agreed, and yes. Changed in **`refactor: address review nits on the session and registry`**.

`SyncEngine` now owns its own observation task over `registry.serverIDs` using `withObservationTracking`, rather than being handed a callback:

```swift
withObservationTracking {
  _ = self?.registry.serverIDs
} onChange: { … }
```

with its own `knownServerIDs` for "which servers have I already given an initial sync" — which is engine state, not a second copy of lifecycle. The registry stays the sole observer of the connections table; the engine observes the set the registry publishes.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

AppShared/Sources/AppShared/Networking/ServerSession.swift
```diff
@@ -0,0 +1,529 @@
...
+      if let built, built.connection == connection {
+        return built.repository
+      }
+      if built != nil {
+        Logger.sync.info(
+          "Server \(stored.logLabel, privacy: .public) connection changed; rebuilding stack")
+      }
+      let repository = try await makeCachingRepository(
+        for: stored, database: database, manager: manager, mode: mode)
+      built = (connection, repository)
```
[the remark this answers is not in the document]
paulgessinger: Verified and fixed in **`fix: serialize repository preparation per session`**.

`prepareRepository` now holds the in-flight build in a `building` slot. A caller arriving while one is running awaits it and then re-evaluates rather than starting a second — it usually finds `built` already matching and gets the winner's repository, and otherwise builds against config read *after* the wait, so a re-auth landing mid-build isn't overwritten by a stale result. There is no suspension point between the wait loop exiting and the slot assignment, and the session is main-actor isolated, so the slot cannot be claimed twice.

This was reachable specifically because deleting the engine's "skip the active server" guard let a sweep and an `activate` land on the same session.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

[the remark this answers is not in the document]
paulgessinger: Verified and fixed in **`fix: retire in-flight phase work when a session rebuilds its repository`**.

This was a regression introduced by this PR, not a pre-existing issue: `DocumentStore.install()` used to cancel in-flight work on every repository swap. Dropping that was correct for server *switches* (the outgoing server's work is now just an inactive session's work and should keep running) but the same guard also covered *connection changes on the same server*, and that half wasn't replaced.

`prepareRepository` now calls `retirePhases()` when it detects a changed `Connection`, cancelling and clearing the four phase slots. Deliberately **not** the composed `syncTask` slot: `runSync` calls `prepareRepository` before its own phases, so cancelling it there would cancel the task doing the discovering — retiring the phases is enough, as the composed pass then sees them fail out.

The connection comparison is load-bearing here and only works because `StoredConnection.token` reads the Keychain live: a re-auth writes a new token without changing the record, so comparing records rather than `Connection`s would keep serving the rejected token.

<sub>— written by Claude (agent) on @paulgessinger's behalf. Verified against the current branch head; commits are referenced by subject because the branch has been rebased since this review.</sub>

Issue #660, opened 2026-08-10
CachingRepository blocks the main thread on SQLite


paulgessinger:
Split out of #656 (item 2) so it can be scheduled. **Full context and rationale: #656 item 2.**

`CachingRepository` is `@MainActor` and every `database.*` call it makes is a synchronous blocking `writer.read` / `writer.write` with no `await` hop. `syncElements()` fans its eight element collections out concurrently in a `withThrowingTaskGroup`, but because the class is main-actor-isolated every write still lands on the main thread — eleven write transactions per sync (8 × `replaceElements`, each a delete-all plus N upserts, plus `setUISettings`, `setServerConfiguration`, `setLastSyncAt`). The cache reads block it too.

With `busy_timeout = 5000`, a ShareExtension holding the writer lock can stall the main thread for up to five seconds. On a server with a few thousand users/correspondents it is a visible hitch on every foreground.

This is a latency/jank regression, not a correctness or data-loss one. It was held mid-stack because applying it would churn ~60 upstack commits.

**Blocked on: the offline/persistence stack landing** (#598 → #600 → #617 → #618 → #620 → #627).

**The point of this issue is to settle the isolation story once** — make the repository non-`MainActor` and hop, versus give `Persistence` async accessors — rather than patching individual call sites. `SyncEngine` is also `@MainActor`, so it needs to be part of whatever decision is made.

paulgessinger:
🤖 _Posted by Claude on Paul's behalf, from manual testing of #600._

## Observed symptom: pull-to-refresh scroll hitch

A concrete, user-visible instance of the jank this issue describes.

**Observed** (Paul, on device, at #600's head): pulling to refresh the document list, at the moment the pull passes the trigger threshold, the scroll stops tracking the finger for a beat and then snaps to catch up to where the finger actually is. The refresh indicator hitches with it. Nothing on screen changes during the refresh — no new document version, no content swap — so this is not a content-churn repaint.

**Verified by reading the code:**

- `CachingRepository` is `@MainActor` (`CachingRepository.swift:98`) and its database calls are synchronous, not `async` — e.g. `writeQueryPage` is `throws(DatabaseError)` over a blocking `writer.write` (`Database+Documents.swift:50`). The SQLite work runs on the main thread, so the gesture cannot be serviced while it runs.
[...]

Issue #656, opened 2026-08-09
Offline/persistence stack: deferred, held, and unverified items from the stacked review


paulgessinger:
Tracking the remaining debt and manual verification from the offline/persistence PR stack:

#588 → #590 → #591 → #592 → #595 → #596 → #597 → #598 → #600 → #617 → #618 → #620 → ~~#627~~

**Every PR in the stack has now landed or been closed.** Twelve merged (#588–#620, plus the follow-on #685); **#627 was closed unmerged** on 2026-09-05, which orphaned three items below; all three have since been re-homed — see the 2026-09-05 fact-check notes. What remains is the table below, no longer gated on any open PR of the stack. Detailed implementation history belongs in the originating PRs and linked issues. This issue is the compact inventory.

## Open work

| # | Item | Status / owner |
|---|---|---|
| 1 | Reclaim superseded `ContentStore` versions and blobs for deleted documents | #687; PR #710 |
| 2 | `CachingRepository` performs synchronous SQLite work on `@MainActor` | Landed in #699 (merged 2026-09-06); #660 closed |
| 3 | Callers joining an in-flight download receive no progress callbacks | #688; PR #711 |
| 4 | Automated coverage for forbidden → `ResourceForbidden` paging behavior | Landed in #702 (merged 2026-09-06) |
| 5 | Exercise the merged re-auth flows manually | #693 |
| 6 | Verify that editing extra headers preserves live cached elements | #693 |
| 7 | Test pull-to-refresh during an active document load | Code landed in #620; manual verification remains in #693 |
| 9 | Narrow collection/document observations so writes for another server do not re-emit | Landed in #700 (merged 2026-09-06) |
| 10 | Remove stale `Stage N` comments | Landed in #703 (merged 2026-09-06) |
| 11 | Decide whether to remove unreachable `CachingRepositoryError.cacheMiss` | Resolved on PR #699: it is reachable (cold cache before the first element sync, and the "no caching backend" path); kept, with reachability documented on the case |
| 14 | Land the #596 review fix (`friendlyNameTask` → `Subscription`) | Landed in #703 (merged 2026-09-06) |
| 15 | Give sync failures appropriate logging and a user-visible surface | #663 |
| 17 | Distinguish device-offline from server-unreachable failures | #667 |
| 20 | Decide when R3δ document refreshes should mark cached query ordering stale | #689 |
| 22 | Avoid a permissions toast from notes prefetch on every refresh | Landed in #620 |
[...]

PR #700, opened 2026-09-05
perf(persistence): stop another server's writes from re-emitting observations

### LINKED ISSUE #696: Narrow per-server observations so another server's writes don't re-emit
Parent tracker: #656 (item 9).

## Problem

The typed live queries in `Persistence/Sources/Persistence/Database+Observe.swift` filter their *fetch* by `server_id`, but GRDB's `ValueObservation` tracks the whole table region. A write to any server's rows therefore re-fires every subscriber, which then re-maps and re-publishes an identical array for the servers that did not change.

The file says so itself:

> Scope: the observation tracks the whole table region, so a write to *another* server's rows re-fires this stream with an identical array. That is harmless while one server is active and not worth narrowing.

That rationale was written when only the active server was ever written. It no longer holds: since #618 and #685, `SyncEngine`'s inactive sweep keeps *every* configured server's cache warm, so background writes for server B now re-emit on the foreground observations for server A while the user is looking at them.

Affected observations, all of which take a `serverID` and are re-fired by unrelated servers' writes:

- `observeElements(_:serverID:)`
- `observeUISettings(serverID:)`
- `observeServerConfiguration(serverID:)`
- `observeDocumentPrefix(queryKey:serverID:limit:)`
- `observeQueryStatus(queryKey:serverID:)`
- `observeDocument(serverID:id:)`
- `observeDocumentCount(serverID:)`

## Impact

Wasted work rather than incorrect data — each spurious emission re-runs the fetch and mapping, and `observeDocumentPrefix` re-runs the `query_order ⋈ document` join up to the scrolled-to `limit`. The cost scales with the number of configured servers and the sweep's write rate, and lands on whatever actor the consumer publishes to.

## Desired outcome

- A write for one server does not re-emit observations scoped to a different server.
- Identical-value emissions do not reach consumers.
- Replace the "not worth narrowing" note in `Database+Observe.swift` with whatever ends up being true.
- Automated coverage: subscribe for server A, write for server B, assert no emission.

Mechanism is deliberately unprescribed — region narrowing, `removeDuplicates`, or both.

paulgessinger:
Closes #696. Part of #656 item 9.

## Why not region narrowing

Checked against GRDB 7.11.1's source rather than assumed: a `DatabaseRegion` is `(table, columns, rowIDs)`, and the rowID set is narrowed only when the filter is on the rowid column itself (`SQLQueryGenerator.optimizedSelectedRegion` → `identifyingRowIDs`). These tables are keyed `(server_id, …)` with an implicit rowid, so a `server_id` predicate is not expressible as a region. A hand-built "rowids of server A" region would also be resolved once at observation start and go stale on the next insert, trading wasted work for stale UI. This applies equally to all seven observations; the join in `observeDocumentPrefix` is not specially worse.

## What changed

All seven observations fetch records, `removeDuplicates()`, and only then map to domain values. A foreign write still costs one SQL fetch (GRDB's floor) but no longer reaches the consumer, and for five of the seven the record→domain mapping is skipped as well. `removeDuplicates()` never suppresses the first value, so the "initial hydrate + live updates" contract is unchanged.

- `observeElements` gains `R: ElementRecord & Equatable` (every conformer already was).
- `UISettingsRecord` and its payload structs gain `Equatable`.
- The "not worth narrowing" header note is replaced with the explanation above, naming the GRDB symbols so it can be re-verified.

## Tests

Nine cases: subscribe for server A, write for server B, then write for A and assert the next emission is A's. All nine fail against the pre-change implementation (negative control run). The helper sleeps 250 ms between the foreign and sentinel writes so GRDB cannot coalesce them into one notification; if that ever flakes on CI, raising it is the fix, not removing it.

## Verified

Persistence 143 tests pass; other three package suites pass; simulator build succeeded (the `observeElements` signature changed); swift-format clean.

## Merge order

Touches the same test files as #660's branch; whichever lands second needs a trivial `try` → `try await` rebase.

## Manual test checklist

- [x] Two servers configured. While viewing server A's document list, trigger a sync of server B (switch to B and back, or wait for the inactive sweep): A's list does not flicker or lose scroll position.
- [x] On the active server, create a tag: it appears in the tag list without a refresh (live updates still work).
- [x] Switch filters and saved views: the list still repaints.

### DIFF
--- a/Persistence/Sources/Persistence/Database+Observe.swift
+++ b/Persistence/Sources/Persistence/Database+Observe.swift
@@ -15,29 +15,59 @@ import GRDB
 /// assigns it directly and never re-reads the database to find out what
 /// changed.
 ///
-/// Scope: the observation tracks the whole table region, so a write to
-/// *another* server's rows re-fires this stream with an identical array. That
-/// is harmless while one server is active and not worth narrowing.
+/// ## Scope: why these streams stay quiet for other servers
+///
+/// A GRDB observation tracks a *database region*, and a region is only
+/// `(table, columns, rowIDs)`. The rowID set is narrowed solely when the
+/// request filters on the rowid itself (`SQLQueryGenerator.optimizedSelectedRegion`
+/// → `SQLExpression.identifyingRowIDs`), and none of these tables is keyed by
+/// rowid — they are keyed `(server_id, …)`. A `server_id` predicate is
+/// therefore *not expressible* as a region, and there is no region narrowing to
+/// be had: a write for another server still wakes every observation on that
+/// table and re-runs its fetch. That matters now that the inactive-server
+/// sweep keeps every configured server's cache warm, not just the active one.
+///
+/// What such a write no longer does is reach the consumer. Every observation
+/// below fetches **records**, applies `removeDuplicates()`, and only then maps
+/// to domain values, so for a server whose rows did not change the fetched
+/// records compare equal, the stream stays silent, and neither the
+/// record → domain mapping nor the consumer's re-publish (main-actor hop,
+/// SwiftUI invalidation, store re-derivation) runs. The residual cost of a
+/// foreign write is one SQL fetch per live observation — the floor GRDB's
+/// region granularity allows.
+///
+/// `removeDuplicates()` never suppresses the initial value: the first emission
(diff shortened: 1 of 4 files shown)


paulgessinger:
@codex

PR #702, opened 2026-09-05
fix: one server projection per server, so a stale observation can't overwrite the current one

### LINKED ISSUE #661: Cover PageCursor's forbidden → ResourceForbidden rewrite with a test
Split out of #656 (item 4) so it can be scheduled. **Full context: #656 item 4.**

The highest-severity finding of the stacked review has no automated coverage. `PageCursor` rewrites `RequestError.forbidden` into `ResourceForbidden<Element>`, which is not a `RequestError` — so the original `catch let error as RequestError where isSkippable` never matched, and a non-superuser lacking `view` permission on `user`/`group` threw out of the task group and permanently killed the entire element sync.

The fix shipped in #597 (`isSkippable(_ error: any Error)` with `error is any ResourceForbiddenError`), but `CachingRepository` lives in `AppShared`, which has no test target. `PageCursorTest` covers paging, cancellation and URL fixing — never the `forbidden` → `ResourceForbidden` rewrite.

**Actionable now, independently of the stack.** `PageCursor.swift` and `PageCursorTest.swift` are untouched by all six open PRs, so a test asserting `RequestError.forbidden` → `ResourceForbidden<Element>` lands on `main` today with no rebase cost.

Scope note: this covers the *rewrite* half only. `isSkippable` itself stays uncovered while it lives in `AppShared` — closing that would mean moving it somewhere testable, which is a separate call.

### LINKED ISSUE #691: Prevent superseded server observations from overwriting current state
Parent tracker: #656 (item 34).

## Problem

When `DocumentStore` switches servers, it cancels the old observation tasks and immediately starts new ones. A value already yielded by an old stream can still be assigned after the new stream starts, allowing state from the previous server to overwrite current state.

`libraryCoverageAt` is especially visible because it may not update again for a long time. Other observed server-scoped properties should be audited for the same race.

## Acceptance criteria

- Values emitted by a superseded server/session cannot mutate the current store.
- Audit every observation restarted during a connection change, not only `libraryCoverageAt`.
- Add a regression test that delays an old observation until after the active server changes.

A generation token is one possible implementation, but the required invariant is session-safe assignment rather than a particular mechanism.

paulgessinger:
Closes #691 and closes #661. Part of #656 (items 34 and 4).

## #661: `PageCursor` forbidden → `ResourceForbidden` test

Two tests in `PageCursorTest`, tagged `.bug(id: 661)`: a fetch throwing `RequestError.forbidden` makes `nextPage()` throw `ResourceForbidden<Int>` (asserting the concrete type, the `ResourceForbiddenError` existential, that the server detail survives, and that it is **not** a `RequestError`, the property that made the old `catch let error as RequestError` never match); and that the rewrite on a later page propagates through `collectAll()`. `isSkippable` lives in AppShared and stays uncovered.

## #691: one projection object per server

The race: every observation loop suspends in `for try await`, then hops back to the main actor to assign. A switch cancels the old loops and starts new ones, but an old loop that already has a value in hand resumes *after* the switch and assigns the previous server's value. `libraryCoverageAt` made it visible because it may not update again for a long time.

Rather than guarding every assignment, the server-scoped projection is now an object with one server for its whole life, replaced wholesale on a switch. A stale value from an old loop can only land in an old object that nothing displays, so the bug class is gone rather than checked for.

- `ElementStore` → **`ServerProjection`** (`git mv`). `init(database:serverID:)` starts all thirteen loops; `deinit` cancels them. `repoint`, `reset`, `clearProjection` and the replace-tasks dance are deleted; a fresh instance is born empty, which is what the old synchronous clear was for.
- It now also owns the three server-scoped values that used to be observed on `DocumentStore` itself and restarted on every switch: `libraryCoverageAt`, `syncErrors`, `cachedDocumentCount`.
- `DocumentStore.projection` is a tracked `public private(set) var ServerProjection?`; every delegate (`store.tags` etc.) reads through it with empty fallbacks for the logged-out case. Both hops are observation-tracked, so views repaint on the swap. Nothing outside `DocumentStore` held the old instance (one preview helper, updated).
- The two sites that are not observations and never restart, `fetchTasks()` and the `lastSyncError` writes in `sync()`, pin the session before their await and compare identity after it. No helper type.

Net −14 lines. The wiring stays untested: AppShared has no test target and the store needs a registry plus a GRDB stack to construct; nothing was faked to cover it.

## Behaviour notes

- On a switch, coverage timestamp, sync errors and document count now blank until the new server's first emission, the same as the element collections already did. Visible on the Offline & Sync screen for one runloop hop.
- Re-activating the *same* connection (e.g. after editing extra headers) rebuilds the projection, equivalent to the old `repoint`, which also cleared.
- Comment-only mentions of "ElementStore" remain in `CachingRepository`, `ConnectionsView` and `ShareView`; worth a follow-up sweep.

## Verified

All four package suites pass; `just generate` + simulator build succeeded; `just lint` clean.

## Merge order

`refreshUISettings` becomes `async` under #699 (it reads a cache table). Whichever of #699 and this lands second needs that two-line rebase; the integrated tree of both has been built locally.

## Manual test checklist

- [x] Switch between two servers: tags, correspondents and saved views change immediately with nothing left over from the previous server.
- [x] After switching servers, Offline & Sync's document count and "last full fill" belong to the new server.
- [x] Remove the active connection: lists clear and nothing stale reappears.
- [x] Edit extra headers on the active connection and save: elements are still present afterwards.

### DIFF
--- a/AppShared/Sources/AppShared/Document/DocumentStore+Preview.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore+Preview.swift
@@ -14,8 +14,8 @@ import Persistence
 
 extension DocumentStore {
   /// A store backed by an in-memory seeded DB, wrapping `wrapped` in a
-  /// `CachingRepository` so the live `ElementStore` projection and the
-  /// write-through mutations behave exactly as in production.
+  /// `CachingRepository` so the live `ServerProjection` and the write-through
+  /// mutations behave exactly as in production.
   ///
   /// The wrapped repository's element data is copied into the DB by a one-shot
   /// `sync()` (previews are live, so it lands a beat after first render);
@@ -52,7 +52,7 @@ extension DocumentStore {
     // ownership path production does, and there is no `init(repository:)` for
     // production code to reach for.
     let store = DocumentStore(session: ServerSession(serverID: serverID, repository: caching))
-    store.elementStore.refreshUISettings(from: database, serverID: serverID)
+    store.projection?.refreshUISettings()
     Task { try? await store.sync() }
     return store
   }
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -22,28 +22,34 @@ public final class DocumentStore: Sendable {
 
   public private(set) var tasks: [PaperlessTask] = []
 
-  /// The live element projection (DB → typed `ValueObservation`). The store owns
-  /// it and re-exposes its collections through the computed delegates below, so
-  /// `store.tags` etc. keep working and stay reactive (a view reading
-  /// `store.tags` tracks `ElementStore.tags` through the getter). The reference
-  /// is stable across its lifetime — only its contents change — so it is
-  /// `@ObservationIgnored`; the inner `@Observable` does the tracking.
-  @ObservationIgnored
-  public let elementStore = ElementStore()
+  /// The live projection of the *active* server (DB → typed `ValueObservation`),
+  /// or `nil` when there is no server to project (before login, or a repository
+  /// that fronts no DB). The store owns it and re-exposes it through the
+  /// read-only computed delegates below, so `store.tags` etc. keep working.
+  ///
+  /// Deliberately a tracked stored property, and deliberately *replaced* rather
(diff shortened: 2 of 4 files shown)


paulgessinger:
@codex

paulgessinger:
@codex review

PR #703, opened 2026-09-05
fix: friendly-name Subscription and stale Stage N comments

### LINKED ISSUE #662: Sweep stale `Stage N` references from code comments
Split out of #656 (item 10) so it can be scheduled. **Full context and the per-file breakdown: #656 item 10.**

29 `Stage N` references in `*.swift` at the stack tip: 8 already on `main`, 21 added by the open stack across 17 files. They point at stages of a planning document that is not in the repository, so they are unresolvable for any reader.

**Partially actionable now.** 5 of the 8 on `main` sit in files no open PR touches — `Broadcaster`, `ApiSavedView`, `DocumentModel`, `CustomFieldModel`, `ShareLinkModel` — and can be swept today with no conflict. The other 3 (`Database+Connections` ×2, `MetadataModel`) are in files the stack edits; leave those and the stack's own 21 until it lands.

Two caveats when sweeping:

- One reference is a legitimate cross-document pointer rather than a bare stage tag (`CachingRepository.swift` cites the plan's Stage 8 by name).
- Three point at stages that do not exist in the tree yet (`QueryKey.swift` → "Stage 14's pinned Downloaded list", `Database+Connections.swift` → "Stage 7 will add…").

Those want rewording, not deletion.

paulgessinger:
Two follow-ups from the offline/persistence stack that were riding on #627 (closed unmerged) and are rewritten here against current `main`.

- **`friendlyNameTask` → `Subscription`** (#656 item 14, the #596 review fix): the friendly-name observer in `swift_paperlessApp` was a bare `Task` held in `@State`; it now uses `Persistence.Subscription`, which cancels on `deinit`, so the observer cannot outlive the view.
- **Drop stale `Stage N` comments** (closes #662): 24 references across the packages rewritten to say what they mean in present tense rather than by stage number. No code change in that commit.

Verified: simulator build clean; Common/DataModel/Networking/Persistence suites pass.

Refs #656.

## Manual test checklist

- [x] Log in to a server whose paperless-ngx has an app title configured: the connection shows that title as its name, and it updates if the title changes on the server and the app is refreshed.
- [x] Open and close the document list a few times: no regression (the observer is now released with the view).

### DIFF
--- a/AppShared/Sources/AppShared/Document/DocumentStore.swift
+++ b/AppShared/Sources/AppShared/Document/DocumentStore.swift
@@ -762,7 +762,7 @@ public final class DocumentStore: Sendable {
   }
 }
 
-// MARK: - Document cache surface (Stage 8)
+// MARK: - Document cache surface
 
 /// The GRDB-free surface the document list and detail views reach for. Each
 /// method resolves the active `CachingBackend` (the production/preview stack) and
--- a/AppShared/Sources/AppShared/Model/OfflineBrowsingMode.swift
+++ b/AppShared/Sources/AppShared/Model/OfflineBrowsingMode.swift
@@ -9,7 +9,7 @@ import Foundation
 /// How aggressively the offline cache fills a given server's documents.
 ///
 /// - `recentlyBrowsed`: only queries the user actually opens are cached
-///   (Stage 8 on-open fill).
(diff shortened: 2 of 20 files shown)


paulgessinger:
@codex

Issue #679, opened 2026-08-22
Thumbnail traffic is neither counted by the transfer meter nor gated by Sync over cellular


paulgessinger:
The data-transfer meter added on #617 reports API JSON only. Thumbnails are the app's other
automatic network consumer, and they are counted by nothing and gated by nothing — so the
number the Offline & Sync screen shows is smaller than what the app actually moves, by an
amount nobody can currently state.

## What the app fetches on its own today

`DocumentListViewModel.prefetchThumbnails` runs from inside the document observation, with
the whole observed prefix, deduped against `prefetchedIds`. That window is `initialLimit =
250`, growing by `widenStep = 250` whenever scrolling comes within `fetchMargin = 25` of the
end, and resetting on `subscribe(to:resettingWindow: true)` — i.e. on load and on every
filter or saved-view change. So opening a list starts ~250 thumbnail downloads against ~10
visible rows, and each widen or filter switch starts another 250.

**This aggressiveness is deliberate and stays.** It's what fixed thumbnail pop-in and scroll
stutter; nothing here proposes reducing it.

Alongside it, each rendered cell does two loads: `DocumentCell`'s `.task(id:
document.versionQueryID)` fetches the resized thumbnail, then calls
`store.preloadThumbnail(for:)`, a second `.high`-priority *unprocessed* load of the same URL
for the detail view. Nuke's on-disk `DataCache` (app group, `Caches/Nuke`) means repeats
don't refetch.

## Why none of it is counted

The meter has exactly one feeder: `NetworkTransfer.record(bytes:category:)` inside
`ApiRepository.fetchData`. The image pipeline (`DocumentStore.makeImagePipeline`) hands Nuke
a `DataLoader`, which owns its own `URLSession` — only the *delegate* is shared, for
client-cert/TLS challenges. Nothing on that path reaches `fetchData`. Streamed file
downloads (`streamDownload`) are excluded for the same reason.

`TransferStats.swift`'s header comment states this deliberately. The screen doesn't: the
Data section's footer says only "since <date>" — which is the labelling half of #656 item 33.

## Why none of it is gated

`NetworkMonitor.allowsProactiveSync(syncOverCellular:)` covers `fillLibraryIfEnabled` and
`fillDocumentDetailsIfEnabled` — the two proactive JSON sweeps — and nothing else. Nuke's
`DataLoader()` takes the default `URLSessionConfiguration`, so `allowsConstrainedNetworkAccess`
[...]

PR #712, opened 2026-09-06
fix: separate interactive list traffic from proactive fills in the transfer meter

### LINKED ISSUE #694: Correct transfer-meter attribution and scope labels
Parent tracker: #656 (item 33).

## Problem

The transfer screen overstates the precision of its categories:

- Interactive list fills and proactive library fills share `fillQuery` and are both recorded as `.fill`, so user-initiated traffic appears under “Library fill.”
- Labels such as “Data transferred” and “Total” read as comprehensive even though some traffic is intentionally outside this meter.

Thumbnail accounting is tracked separately in #679. This issue owns correct attribution and labeling for traffic the meter already observes.

## Acceptance criteria

- Callers pass the correct transfer category into the shared fill path.
- Interactive list traffic is distinguishable from proactive/background fill traffic.
- Labels or explanatory copy accurately state the meter's scope.
- Tests cover transfer-category attribution where practical.

paulgessinger:
### Correct transfer-meter attribution and scope labels

`fillQuery` serves both an interactive list open and the proactive library sweep, and hardcoded `.fill` for both, so traffic the user asked for was reported as background sync, the one distinction the meter exists to make.

The category is now a parameter of the shared path, since the paging runs detached and inherits no task-local. `DocumentStore.fillDocumentQuery` (the list UI's only entry point) passes a new `.list`; `fillLibrary` passes `.fill`, which now means proactive fills only. The Offline & Sync screen gains a "Browsing lists" row.

Totals persist keyed by raw value and drop keys they can't decode, so the added case is compatible in both directions. Existing `.fill` totals keep displaying, blended, until a reset. A test pins the raw values so a rename can't silently zero a category's history.

The screen also claimed more than it counts: thumbnails and streamed downloads take other paths (see #679), so "Data transferred" and "Total" read as comprehensive. The section header is now "API data transferred", with a footer naming what is and isn't included. The old header key was replaced rather than edited so Crowdin gets a fresh string instead of stale translations.

Tests cover the default category, nested-scope precedence, and raw-value stability.

Closes #694. Part of #656 (item 33).

### Manual test checklist

- [ ] Settings → Offline & Sync: header reads "API data transferred", footer shows "Since …" plus the scope sentence naming thumbnails and downloaded files as excluded.
- [ ] Note the per-category figures, open the document list and switch to a saved view, return: a "Browsing lists" row appears or grows, and "Library fill" does not grow from that action.
- [ ] With "Entire library" on and the background fill running, "Library fill" grows while "Browsing lists" stays put.
- [ ] Tap Reset statistics: all rows clear, "Since" updates to now, and the next list open starts "Browsing lists" from zero.


paulgessinger:
@codex

PR #721, opened 2026-09-13
feat: count thumbnail and document download traffic in the transfer meter

### LINKED ISSUE #679: Thumbnail traffic is neither counted by the transfer meter nor gated by Sync over cellular

paulgessinger:
### Count thumbnail and document download traffic in the transfer meter

Thumbnails and document file downloads are the app's other network consumers (the list prefetch alone starts ~250 thumbnail loads per list open, and a single PDF can outweigh a lot of JSON), but the transfer meter only had one feeder, inside `ApiRepository.fetchData`. Nuke's `DataLoader` owns its own `URLSession`, and document downloads stream through `URLSession.getDownload`, so none of that traffic was counted, and the Offline & Sync numbers were smaller than what the app actually moves by an unknown amount.

**Thumbnails.** `PaperlessURLSessionDelegate` now implements `urlSession(_:task:didFinishCollecting:)` and reports bytes through an optional `onTransfer` closure. `ApiRepository` derives a second, recording copy of its delegate (`recordingTransfers(_:)`, reusing the already-loaded client-certificate credential rather than hitting the keychain again) and hands that one to the image pipeline. Nuke's `DataLoader` already forwards the metrics callback to its assigned delegate, so no proxying is needed. The API session's own delegate is untouched, and API requests keep their per-task delegate, so nothing is double-counted.

`Repository.delegate` is renamed to `imageSessionDelegate`: its only consumer is `makeImagePipeline`, and a generically named property that now returns a *recording* delegate would be a trap for the next caller.

**Document files.** `getDownload` gets the same `onTransfer` parameter as `getData`, and both download paths (`streamDownload` into the `ContentStore`, and `fetchToTemp` when there's no staleness stamp) pass `ApiRepository.recordDocumentTransfer`, which files the bytes under a new fixed `.documents` category. Fixed rather than task-local, because the callback runs outside the caller's task. Coalesced callers share one task, so a download joined by two screens is recorded once. A document served from the `ContentStore` makes no request and counts nothing. Uploads already went through `fetchData` and were counted before this PR.

The byte summation over `URLSessionTaskMetrics` is one shared helper used by every feeder.

**UI.** New `TransferCategory.thumbnails` and `.documents` rows ("Thumbnails", "Document files"). The section header was "API data transferred", which is no longer accurate, so it becomes "Data transferred". #712's footer said thumbnails and downloaded files were excluded; it is replaced with a new key that lists what is counted and notes that opening a document from the offline cache uses no data. Both replaced strings use new keys, so Crowdin gets fresh strings rather than stale translations.

Gating thumbnails on cellular / Low Data Mode is deliberately not part of this: the 250-row prefetch window stays, and the gating decision is meant to be made with these numbers in hand.

Tests cover the summation helper, the delegate's recording seam, that `imageSessionDelegate` records under `.thumbnails` and `recordDocumentTransfer` under `.documents` even inside a different task-local category, and that `getDownload` invokes `onTransfer` for a download task (against a `URLProtocol` stub, which reports no wire bytes, so the byte counts themselves can only be checked by hand). `URLSessionTaskMetrics` has no usable initializer.

Closes #679. Part of #656 (item 36).

### Manual test checklist

- [x] Settings → Offline & Sync → Reset statistics, then scroll a document list: a "Thumbnails" row appears and grows.
- [x] Open a document not yet cached: a "Document files" row appears, roughly the file's size. Open it again: the row doesn't grow.
- [x] Opening a list does not inflate "Browsing lists" or "Other" beyond the JSON traffic.
- [x] The new header and footer text read correctly.
- [x] On a server that requires a client certificate, thumbnails and document downloads still load (the pipeline now uses the derived delegate).

https://claude.ai/code/session_0137Gp2EU4pSXez1ApezqnZk


paulgessinger:
@codex

Issue #678, opened 2026-08-22
Recently browsed cache cap is only enforced at the downgrade transition, not ongoing


paulgessinger:
`reclaimAfterDowngrade`'s truncate-to-`recentlyBrowsedDefaultListCap` (200) + prune only
ever runs once, at the moment a server flips *Entire library* → *Recently browsed*. Nothing
re-applies it afterward.

`fillQuery` (the function behind opening *any* document list) eager-fills that list in
full, uncapped, regardless of `OfflineBrowsingMode` — deliberately, per its own doc comment
(no on-scroll fetch trigger exists yet, so a cap there would make anything past it
unreachable even online). The default list is what's on screen at every launch, so every
relaunch calls `fillQuery(filter: .default)` on it, background-paging the *entire* list back
to disk. Observed: downgrade to *Recently browsed*, let the reclaim GC settle at 200,
reopen the app — cached documents climbs right back toward the server's full count.

This is payload the app already fetches at list-view time regardless (`fillQuery`'s page
reads are the document JSON only — no PDF/thumbnail bytes; those are still fetched only on
open), so it isn't a new network cost. It's a **local storage** cap that only ever holds for
as long as nobody looks at the list.

Proposed direction: don't touch `fillQuery` (preserves full online+offline browsability of
whatever's been opened — capping the fill itself would need the on-scroll fetch trigger
first, or scrolling past the cap would dead-end even online). Instead make the cap
*recurring* rather than one-shot: re-run the existing `reclaimAfterDowngrade` /
`truncateQuery` / `pruneUnreferenced` machinery periodically for any server currently in
*Recently browsed* — e.g. folded into the regular reconcile sweep — so what's kept on disk
settles back down to the cap sometime after a viewing session, rather than never again.

Open question: cadence/trigger (every reconcile? throttled separately?) and whether the
truncate should key off `query_meta.filled_at` (LRU-ish, keep the most-recently-filled 200)
rather than blindly `keepingFirst` position order, so re-browsing an older document doesn't
get immediately evicted again on the next sweep.

Not introduced by any single PR in the stack — `fillQuery`'s "every opened view eager-fills
in full" is pre-existing, deliberate behavior; the one-shot-only reclaim is what's
incomplete. Surfaced during manual QA of #617's *Entire library* checklist.

PR #724, opened 2026-09-13
refactor: factor sync single-flight slots into TaskSlot and KeyOwnership

### LINKED ISSUE #716: Factor the repeated single-flight/key-ownership idiom out of the sync stack (instead of consolidating onto dedicated actors)
The sync/persistence stack has accumulated a noticeable amount of hand-written
coordination code, and the recurring question is whether consolidating onto one
or two dedicated actors — making `CachingRepository` and the sync types
non-`@MainActor` — would remove it. This issue records the answer (no, and why)
and the refactor that *does* address the underlying duplication, so the question
does not get re-litigated from scratch.

## Why actor consolidation is not the answer

The sync stack is already in exactly one isolation domain (`@MainActor`:
`DocumentStore`, `ServerSession`, `SyncEngine`, `CachingRepository`), plus a
second that is not hand-written at all (GRDB's serialized writer, reached only
through `Persistence.Database`'s `async` API). In that path there is no lock,
no `nonisolated(unsafe)`, and no cross-domain shared mutable state. Swift 6
strict concurrency is not what generates the code.

What generates it is *logical serialization across suspension points*, which
actor isolation does not provide. Per mechanism:

| Mechanism | Why it exists | Removed by a custom actor? |
|---|---|---|
| Five phase task slots + `didSet` → `isSyncing` (`ServerSession`) | Two callers (store, scheduler) want the work to happen once | No — actors don't dedupe work, only interleaving |
| `activeFills` + `drainFill` (`CachingRepository`) | One writer per `QueryKey`'s `query_order`, held across dozens of network pages | No — actor mutual exclusion ends at the first `await`, and a fill is almost entirely awaits |
| `building` slot in `prepareRepository` | An `await` sits between "is `built` still good?" and committing the answer | No — this is textbook actor *reentrancy*; a custom actor has the identical hazard |
| `retirePhases`, the `try Task.checkCancellation()` backstops | Unstructured tasks inherit isolation but not cancellation | No — structured-concurrency semantics, orthogonal to isolation |
| Two reconcile stamps, throttles, `ReconcileResult` | Policy: a broken server must not be throttled away | No — not concurrency at all |

`@MainActor` *is* the single actor; changing which actor it is changes nothing
about reentrancy, and reentrancy is the whole problem.

The costs of moving off it are real, though: every property the views read
(`isSyncing`, `syncActivities`, `offlineBrowsingMode`, the `ElementStore`
projections) becomes `async`, so a `@MainActor` snapshot/publishing layer has to
be added back — net new code. `@Observable` + SwiftUI wants main-actor
properties. And the guarantee currently leaned on for free — *claiming and
clearing a slot never suspends, so a slot cannot be claimed twice* — would have
to be re-verified across every site during a mechanical port, for no behavioural
gain.

The one genuine benefit of a non-main actor would be moving work off the main
thread, and that is already true: DB work runs on GRDB's queues (item 2, landed
in #699), and decoding/paging runs on the `ApiPagedSource` / `PageCursor` actors.
The main actor does bookkeeping, not work.

## What the actual duplication is

Not isolation — one coordination idiom, written out by hand ten times:

```swift
let task = Task { … }
slot = task
defer { if slot == task { slot = nil } }   // "retract only our own"
return await task.value
```

Sites (identity-check line numbers as of 3132dd90d): `ServerSession`
`elementSyncTask`, `reconcileTask`, `libraryFillTask`, `detailFillTask`,
`syncTask`, and `building` (twice, in `prepareRepository`);
`CachingRepository` `activeFills` (three times: the fill registration, the
`drainFill` retraction, and the membership-sweep rewrite). Each carries its own
comment re-deriving why the equality check is needed.

The recent race fixes — 3132dd90d, 813a7460e, 3e49b60ca — are each a bug in an
*instance* of that idiom, not in the idea. And the abstraction already exists:
`Common/SingleFlight.swift` (keyed join-or-start, correct retraction, progress
fan-out to late joiners, unit-tested in `CommonTests`) is currently used only by
`ApiRepository`'s download path.

## Proposed work

1. **A `PhaseSlot<Value>` (or keyed single-flight) for `ServerSession`'s five
   phase slots.** Join-or-start, retract-if-ours, and cancel-all (for
   `retirePhases`) live in one tested place. `isSyncing` becomes "any slot
   occupied" instead of a `didSet` mirror across five properties.
2. **A named `KeyOwnership` type for `activeFills`**, with `claim` / `drain` /
   `isOwned` as an explicit "exclusive writer of this key, across awaits"
   primitive. That is the concept both recent `query_order` race fixes were
   really about; naming it makes the next such question a compile-time one.
3. **Fold the `Task.checkCancellation()` backstop into those two types** — it is
   already at every relevant boundary, just spelled out inline each time.

Behaviour-preserving, `@MainActor`-scoped, and mostly mechanical. Both new types
should land where a `swift test` target can reach them (`Common`, most likely) —
`AppShared` has none, which is the same constraint that keeps `SyncPlan` in
`DataModel`.

Explicitly **not** in scope: making `CachingRepository`, `ServerSession`, or
`SyncEngine` non-`@MainActor`.

Part of #656.

paulgessinger:
### Factor the single-flight / key-ownership idiom into TaskSlot and KeyOwnership

The sync stack hand-wrote one coordination idiom about ten times (start a task, park it in a slot, clear the slot only if it still holds *our* task, await it), each copy re-deriving its own justification, and the recent race fixes were bugs in individual copies. As #716 records, the fix is naming that idiom, not moving `ServerSession`/`CachingRepository` off `@MainActor`; this PR stays entirely `@MainActor`-scoped and behaviour-preserving.

**New in `Common`** (so `swift test` reaches them):
- `TaskSlot<Success, Failure>` (`@MainActor @Observable`): `joinOrStart(_:)`, and `joinOrStart(ifIdle:)` whose decide closure is synchronous, so the compiler rules out a suspension between "slot is empty" and "claim it". `retire()` cancels and empties at once; `waitUntilIdle()` serves the repository build; it clears only its own task, exposes observable `isOccupied`, and cancels leftovers on deinit.
- `KeyOwnership<Key>` (`@MainActor`): `claim`, `release(_:ifOwnedBy:)`, `drain`, `isOwned`, `ownedKeys`, and `withOwnership(of:perform:)`, which runs the write unstructured, returns `false` if it was drained, then calls `Task.checkCancellation()` so a cancelled *caller* never records drained work as done. That absorbs the inline cancellation backstops.

Both document the "claiming never suspends" invariant, and why a main-actor caller reaches the claim without suspending.

**Adopted in** `ServerSession`: the five phase slots, the repository `building` slot, `retirePhases`, `invalidate`, `deinit`; `isSyncing` is now computed from the slots instead of a `didSet` mirror. And in `CachingRepository`: the fill claim and watcher release, `drainFill`, `isFilling`, `replaceQueryOrderOwningKey`, and the unreachable-query sweep. Left hand-written: the fill tasks' own `checkCancellation()` result checks and per-page loop checks, and `runReconcile`'s per-sweep policy. Those are not slot or ownership boundaries.

**Why not `SingleFlight`.** It is lock-based and isolation-agnostic, clears its entry from inside the operation (so `isSyncing` would drop before the caller resumes), has no cancel (so no `retirePhases`), isn't observable, and its policy is join-the-owner where `KeyOwnership`'s is drain-and-replace. Both new types say how they differ.

**Behaviour preservation.** Join checks still precede the per-phase guards, and `lastReconcileAttempt` is still stamped only when a pass starts. Tasks are still created from the caller's main-actor context (inheriting priority and `NetworkTransfer.$category`) without inheriting caller cancellation. The prior sync race fixes each have a test: a sweep owns exactly the keys it saw and an arriving fill drains it (813a7460e); a cancelled sweep is not reported as success (3e49b60ca / 246ae2c35); one repository build at a time without inheriting a failure (f2e7b6bc3); a retired slot starts fresh (3f4087f00); `isSyncing` covers the same five slots (3c6f2d190). A mutation check (breaking clear-only-ours, dropping the cancellation check) fails four of the new tests.

Two observable differences: a sweep that is both cancelled and drained now throws `CancellationError` rather than logging "drained" and returning, which `ServerSession`'s backstop already turned into a cancelled pass, so the recorded outcome is the same. And views reading `isSyncing` may re-evaluate on each slot change rather than only on the combined flip.

**Pre-existing race, not fixed here.** If two `fillQuery` calls for one key drain the same owner concurrently, the second `claim` replaces the first without cancelling it and both write `query_order`. The `claim` doc now states the replace-without-cancel semantics; looping `drain` until free before claiming (as `waitUntilIdle` does) would close it.

**Overlap with #678.** Both PRs rewrite the sweep in `collectUnreachableQueries`; whichever lands second moves #678's cap write inside `withOwnership(of: collected.union(capped))` and uses `ownedKeys`. That resolution is built and tested in my local integration merge.

17 new tests in `CommonTests` (93 total pass); the app builds at every commit. Net: ServerSession shrinks by ~100 lines; most of the addition is tests.

Closes #716. Part of #656 (item 40).

### Manual test checklist

- [x] Launch with multiple servers: Offline & Sync shows sync activity starting and settling; the syncing indicator clears when all phases finish.
- [x] Switch servers mid-sync: the previous server's phases retire and the new server syncs fresh (no stuck spinner).
- [x] Open a list while a reconcile sweep is running: the list fills fully (no truncated cached list).
- [x] Edit connection settings (triggers a repository rebuild) during a sync: no duplicate builds, sync resumes.

https://claude.ai/code/session_0137Gp2EU4pSXez1ApezqnZk


paulgessinger:
@codex

